PR #24, opened 2015-01-29
Display an ASCII PIN matrix

np:
This is a suggestion of usable interface to enter the PIN code using python CLI.

It displays a matrix as follows:

```
 +---+---+---+
 | A | B | C |
 +---+---+---+
 | D | E | F |
 +---+---+---+
 | G | H | I |
 +---+---+---+
```

You can then type the letters which corresponds to what you see on the device (lower and upper case).

I'm wondering if the silent pin entry is of much use in this situation.

### DIFF
--- a/trezorlib/client.py
+++ b/trezorlib/client.py
@@ -7,6 +7,7 @@ import unicodedata
 import mapping
 import json
 import getpass
+import string
 
 import tools
 import messages_pb2 as proto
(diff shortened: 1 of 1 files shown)


prusnak:
This does not make much sense. Entering numbers on keyboard numpad is convenient while using alphabet is not. I agree that prompt might be more verbose but let's not make more complicated than they need to be.

PR #27, opened 2015-03-03
Insight tx fixes

nelisky:
This PR includes two changes I needed to apply for my use:
- The insight_url() call assumes, as the name implies, a url. But I use my own backend which is jsonrpc based so I'm using my own TXAPI class that calls that and then delegates the result to insight_url as a dict object. No sense in redoing the good work already applied to tx_api, right? :)
- Doing Decimal(float_value) is a recipe for disaster as rounding errors are almost certain, at least in OSX and python 2.7.x, which is my work environment. Coercing the float value into a string before making a Decimal out of it completely prevents the rounding issues.

### DIFF
--- a/trezorlib/tx_api.py
+++ b/trezorlib/tx_api.py
@@ -84,7 +84,7 @@ def insight_tx(url, rawdata=False):
 
     for vout in data['vout']:
         o = t.bin_outputs.add()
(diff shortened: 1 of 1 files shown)


prusnak:
Thanks!

PR #28, opened 2015-03-05
Use right URL for bridge and keep-alive connection

jhoenicke:
The bridge is using https with a certificate signed for localback.net.

Use a session object (self.conn) to keep connection alive and
prevent costly ssl handshakes for every call.


prusnak:
Thanks, looks OK.

PR #29, opened 2015-03-05
Verify the localback.net certificate

jhoenicke:
I just noticed that the certificate is signed by startssl and therefore valid.  The reason I added verify=None in the first place was that it didn't work when I had localhost instead of localback.net.


prusnak:
Ah, I thought you had some problems with StartSSL CA. Merging ...

PR #33, opened 2015-04-29
Don't spam the usb.

jhoenicke:
The connected logic was flawed.  After 10 s it would continuously
check if the device is still connected.  Now, we reset the timer
after every check.


slush0:
Thanks!

PR #34, opened 2015-05-10
insight api: use hex code

jhoenicke:
The insight api transaction parser used to look at the assembly to reconstruct the input and output scripts.  This patch changes it to use the hex input directly.  This fixes parsing some instructions with unknown opcodes (e.g. multisig inputs with `0` opcode).


prusnak:
Ah, I reported the issue about missing field here:

https://github.com/bitpay/insight-api/issues/34

But never got informed that it's finally implemented. Cool!

PR #36, opened 2015-05-12
Update hashbang to use /usr/bin/env python

mdance:
Please use the /usr/bin/env python hashbang in order to properly detect alternative python installations (like those installed through homebrew on a mac).


prusnak:
Duplicate of #35

slush0:
It's not a duplicate, it's another file.

slush0:
Thanks!

PR #39, opened 2015-07-08
Add information to README on how to enter scrambled pin

reiven:
@prusnak here i send you a proposal to add some helpful information into README about how to enter scrambled pin using python-trezor

Let you know if you want to correct something


prusnak:
Thanks!

PR #42, opened 2015-09-04
Shamir's secret sharing tool & firmware_update by version number or latest

chrisrico:
I can rebase to remove the whitespace changes and split this up into two pull requests if desired. I just figured I'd see if it's something you're interested in merging before I bothered.


chrisrico:
I just realized I was relying on the pybitcointools package which you don't use.

Should I add the package to the dependencies or is there a better way to convert the number representation of a seed to a bip32 master key? [This](https://github.com/vbuterin/pybitcointools/blob/master/bitcoin/deterministic.py#L135) is the function I'm using currently.

prusnak:
I cherry picked the last commit (firmware update from releases.json) - 6475f98b1ec8d2f1bd9740d516014d9c4dbf5a6b.

About the SSSS, I have a better solution in my pipeline (SSSS directly in TREZOR), so I am not merging that in. But that will take some time until properly implemented.

PR #49, opened 2016-01-07
Added code to make trezor-agent work with keepkey.

bgok:
Would you be interested in taking this pull request? It is needed to make Roman's trezor-agent work for keepkey with minimal changes.

Thanks.

### DIFF
--- a/trezorlib/client.py
+++ b/trezorlib/client.py
@@ -343,7 +343,7 @@ class DebugLinkMixin(object):
 
 class ProtocolMixin(object):
     PRIME_DERIVATION_FLAG = 0x80000000
-    VENDORS = ('bitcointrezor.com',)
+    VENDORS = ('bitcointrezor.com', 'keepkey.com',)
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


prusnak:
Thanks!

dstanchfield:
Anyway we can get this support for KeepKey removed.  This will cause problems with our electrum plugin.

prusnak:
Reverted in 8b7e6058ed8c55bd462c26037c21a73af90d6d86.

dstanchfield:
Thanks!

kyuupichan:
Hmm, maybe I should hard-code the device IDs in Electrum and not import them from the library headers?

prusnak:
Best would be, of course, to create one plugin that would handle both devices. Keepkey's code is just same as TREZOR code with replaced strings anyway.

kyuupichan:
It's not the same actually.

There plugin code is 99% shared in Electrum.

Now if you two could agree on a common single library, that would be the most helpful.  I could then indeed eliminate the plugin separation in Electrum, but not until then.

PR #53, opened 2016-02-19
Note about exclusivity

runn1ng:


prusnak:
Are you sure? I was under the impression that hidapi library will lock the device exclusively and other processes have to wait.

runn1ng:
Yep, but that's on message level. I meant it on "task level", as in, the whole task flow of message exchange will be interrupted with different messages than the processes expect.

runn1ng:
all right, I was wrong, you was right :) you win another one! :cry:

PR #61, opened 2016-05-27
Implement ethereum_sign_tx

Arachnid:


prusnak:
```
from rlp.utils import int_to_big_endian
```

Do you think we'll end up pulling in more stuff from RLP? If not we can consider copying the int_to_big_endian function to skip this dependency? What do you think?

prusnak:
Other option: We already require `ecdsa` in trezorlib/ckd_public.py. I think we can use this method instead: https://github.com/warner/python-ecdsa/blob/04a383c7bc144453593192005b78aba2e989c91e/ecdsa/util.py#L169

Arachnid:
Yup - my follow-up CL also needs rlp.encode in order to generate the encoded transaction.

prusnak:
OK then.

PR #62, opened 2016-05-27
Implement ethereum_send_tx

Arachnid:
This implements everything necessary to sign and send a transaction to the Ethereum network, as long as the user has access to a node running the RPC service. Example:

    $ ./trezorctl ethereum_send_tx --value "100 finney" 0xfB6916095ca1df60bB79Ce92cE3Ea74c37c5d359
    Transaction sent with ID 0xc6233b6f33468ded099dd57f8040392464ddb4d2276c19b078b6a419d41ade26

### DIFF
--- a/trezorctl
+++ b/trezorctl
@@ -11,6 +11,28 @@ from io import BytesIO
 
 from trezorlib.client import TrezorClient, TrezorClientDebug
 
+ether_units = {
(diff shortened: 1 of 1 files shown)


prusnak:
Great! Let's get that in :-)

PR #64, opened 2016-06-27
Update Unit Tests, added SetU2FCounter and debug mem functions

jhoenicke:


slush0:
Awesome, thanks!

slush0:
Out of curiosity, do you run those tests against Trezor firmware or trezor-emu? Does it pass for firmware now?

jhoenicke:
I tested it against Trezor firmware running on real Trezor (my jhoenicke/trezor-mcu@master branch).  For this it passes all tests.  

I think the Trezor master branch still has a small bug with encrypted xprv (master public key is not recomputed after decryption).  Really minor issue, since nobody should be using xprv at all.

PR #67, opened 2016-08-19
Fix ethereum_sign_tx

jhoenicke:
- Workaround bug in rlp.utils.int_to_big_endian
- Command line tool now expects data to be a hexlified string


prusnak:
Thanks!

PR #71, opened 2016-09-18
Avoid repeating keyword argument

sp4ke:
Repeating the keyword argument breaks package build with the following error: "SyntaxError: keyword argument" repeated


slush0:
Thanks!

Issue #76, opened 2016-10-14
Exception: Unexpected magic characters


dmytroleonenko:
latest git trezorlib. Works sometime but later complain about magic characters. firmware 1.4.0, Mac OS X 10.10.5, Python 2.7.12 (homebrew)

```
$ trezor-agent meclo@nowa.cc
Traceback (most recent call last):
  File "/usr/local/bin/trezor-agent", line 11, in <module>
    sys.exit(run_agent())
  File "/usr/local/lib/python2.7/site-packages/trezor_agent/__main__.py", line 124, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/trezor_agent/__main__.py", line 137, in run_agent
    with client_factory(curve=args.ecdsa_curve_name) as conn:
  File "/usr/local/lib/python2.7/site-packages/trezor_agent/client.py", line 20, in __init__
    client_wrapper = loader()
  File "/usr/local/lib/python2.7/site-packages/trezor_agent/factory.py", line 243, in load
    device_list.extend(device)
  File "/usr/local/lib/python2.7/site-packages/trezor_agent/factory.py", line 27, in _load_client
    connection = client_type(hid_transport(d))
  File "/usr/local/lib/python2.7/site-packages/trezorlib/client.py", line 361, in __init__
    self.init_device()
  File "/usr/local/lib/python2.7/site-packages/trezorlib/client.py", line 368, in init_device
    self.features = expect(proto.Features)(self.call)(proto.Initialize())
  File "/usr/local/lib/python2.7/site-packages/trezorlib/client.py", line 85, in wrapped_f
    ret = f(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/trezorlib/client.py", line 98, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/trezorlib/client.py", line 134, in call
    resp = self.call_raw(msg)
  File "/usr/local/lib/python2.7/site-packages/trezorlib/client.py", line 98, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/trezorlib/client.py", line 130, in call_raw
    return self.transport.read_blocking()
  File "/usr/local/lib/python2.7/site-packages/trezorlib/transport.py", line 68, in read_blocking
    data = self._read()
  File "/usr/local/lib/python2.7/site-packages/trezorlib/transport.py", line 128, in _read
[...]

PR #77, opened 2016-10-14
transport_hid: return sorted list of devices

romanz:
This should be deterministic, since dict.values() order is unpredictable.


romanz:
May solve #76 (and https://github.com/romanz/trezor-agent/issues/50 as well)

prusnak:
Thanks!

PR #88, opened 2016-11-30
OpenPGP support

saleemrashid:
This adds support for the initialization of the OpenPGP card

trezor/trezor-mcu#133
trezor/trezor-common#21


trezorctl
```diff
@@ -28,6 +28,8 @@ import json
 import base64
 import tempfile
 from trezorlib import types_pb2 as types
+import time
+import textwrap
```
romanz: I think this import is not required.

saleemrashid: @romanz Was using it for ASCII Armor (read: will be using it again very soon) so I don't really think it's worth rebasing yet

PR #89, opened 2016-12-01
Added support for signing Ethereum transactions without node connection

santiagorp:
I just added some more parameters to the command line client in order to allow custom gas price and manually specifying other parameters (like the nonce) so there is no need to connect to an Ethereum node in order to create the signed transaction.

This is very useful in combination with myetherwallet as you can relay the transaction directly there:

**Instructions:**
Get your trezor ethereum address:
```
trezorctl ethereum_get_address
```

**Send ether**
1. Invoke trezorctl specifying all parameters for offline signing:
 ```
trezorctl ethereum_sign_tx -v '185047 szabo' -g 21000 -t '18 gwei' -i 2 0x45bCD6F98B77E44A812A8CdbBf2642C9dC9403a2
```

Please note the nonce parameter (-i 2) which is the transaction number for that address. You can check the value by adding one to the nonce of the last sent transaction from your trezor address (use a block explorer to check). If is the first transaction you are executing, it should be 0, then 1, 2, 3, etc.
 
2. Enter your PIN and passphrase
3. Copy the Raw Signed Transaction string to the clipboard
4. Go to https://www.myetherwallet.com/#offline-transaction and paste the raw transaction at the end of the page (_Step 3: Send / Publish Transaction_).
5. Click **Send Transaction**

### DIFF
--- a/trezorctl
+++ b/trezorctl
@@ -149,29 +149,59 @@ class Commands(object):
         else:
             value = int(value)
 
+        gas_price = args.gas_price
+        if gas_price:
+            if ' ' in gas_price:
+                gas_price, unit = gas_price.split(' ', 1)
+                if unit.lower() not in ether_units:
+                    raise CallException(types.Failure_Other, "Unrecognized gas price unit %r" % unit)
+                gas_price = int(gas_price) * ether_units[unit.lower()]
+            else:
+                gas_price = int(gas_price)
+
+        gas_limit = args.gas
+        if gas_limit:
+            gas_limit = int(gas_limit)
+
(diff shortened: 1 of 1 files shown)


prusnak:
@Arachnid What do you think?

Arachnid:
LGTM!

prusnak:
@santiagorp could you please remove debug prints on lines 199-204 from your commit?

santiagorp:
I left them so the user could see the details after unit conversion... anyway he can verify on the trezor screen (but the nonce) so I will remove them.

prusnak:
Thanks!

PR #90, opened 2017-01-03
Segwit preparations

jhoenicke:
This adds
 - unit test for signing segwit transactions
 - get_address options for segwit addresses


prusnak:
Thanks!

PR #96, opened 2017-02-11
load_device_by_mnemonic and recovery_device: expand unique mnemonic word prefix matches

mruddy:
related to https://github.com/trezor/python-mnemonic/pull/36

i still need to test these minor changes. putting up now to get the idea published.


prusnak:
I would rather have an extra option that would enable expansion, not have it enabled by default.

mruddy:
OK.
For load_device_by_mnemonic, I think it's pretty easy. Still need to test it.

For recovery_device, I'm stumped for a clean way how to get the expand boolean into TextUIMixin.callback_WordRequest.
I pushed up a new commit, with what I have so far. Could you give me a point in the right direction? Or, maybe we only do this for load_device right now?

mruddy:
Never mind, I think I got it figured out. Just learning python... Will update commit later.

mruddy:
OK, the latest commit should be good. I tested on tails 3 live. If we can get trezorctl, trezorlib, and mnemonic updated in tails 3, this will be useful. Also, before updating tails, there is a Unicode comparison warning in mnemonic.py line 78 that needs to be checked out to make sure it's benign. The warning is unrelated to any of my recent changes. I'll look at that a bit later today.

prusnak:
Thanks! Published python-mnemonic 0.16 and python-trezor 0.7.10 with your changes.

Updating package in Tails requires updating package in Debian first, then it should be updated automatically, I believe.

Not sure what is the policy for updating packages after the release in Debian, but testing/stretch has full freeze now. Better discuss directly with them ...

saleemrashid:
https://release.debian.org/stretch/freeze_policy.html

mruddy:
the unicode warning is the same as was reported previously at https://github.com/trezor/python-mnemonic/issues/29. thanks for updating pypi and the info on debian. i'll check into that later today. gotta run right now.

PR #99, opened 2017-02-20
enhance tools support for reading bootloader from debug link enabled device

mruddy:
enhance tools support for reading bootloader from debug link enabled device.

for this to work, copy the file up one directory so that the module resolves.

Also, the bootloader on my device reports as Loader 1.3.0 and has
```
./mem_read.py 8000000 8000
sha256sum memory.dat 
add28d5faf6de00525bd37d221b195f6ba51e245923e1eadd3db0b7fd9d9caf5 memory.dat
```

This is the disassembly of main that I see, and that I'm having difficulty explaining because it looks like the random32 function is not called and the stack protector is set to zero:
```
     494:       b570            push    {r4, r5, r6, lr}
     496:       4c21            ldr     r4, [pc, #132]  ; (0x51c)
     498:       4b21            ldr     r3, [pc, #132]  ; (0x520)
     49a:       6822            ldr     r2, [r4, #0]
     49c:       b08a            sub     sp, #40 ; 0x28
     49e:       2600            movs    r6, #0
     4a0:       601e            str     r6, [r3, #0]
     4a2:       9209            str     r2, [sp, #36]   ; 0x24
     4a4:       f003 faaa       bl      0x39fc
     4a8:       f003 fbaa       bl      0x3c00
     4ac:       f002 fdf2       bl      0x3094
     4b0:       f7ff ffa8       bl      0x404
```


prusnak:
Thanks!

PR #102, opened 2017-02-26
fix UnboundLocalError cache_file referenced before assignment

mruddy:
fix UnboundLocalError cache_file referenced before assignment. cache_file is only assigned when cache_dir is set


saleemrashid:
I have no context on this code, but from that snippet, I see a check for `cache_file` and a usage of `cachefile`

prusnak:
Good catch!

PR #104, opened 2017-02-28
add tool to aid in rng testing and evaluation

mruddy:
see also
https://github.com/trezor/trezor-mcu/pull/153
https://github.com/trezor/rng-test/pull/1


mruddy:
just fixed a typo in my comment about DEBUG_LINK. it's now use DEBUG_LINK == 1 to disable the button press. the user button press is still on by default in the firmware (which is a good thing).

mruddy:
i'll have to update that comment again now that it's DEBUG_RNG :)

prusnak:
I already fixed that, thanks

PR #107, opened 2017-04-20
Configure Bridge only once per process

tmsrjs:
This saves a lot of time on requests retrieving config_signed.bin and
reconfiguring the bridge for every device enumeration or transport
instantiation.


prusnak:
Good idea, thanks!

PR #112, opened 2017-06-05
Fix client.py and protobuf_json.py for Python 3 support

romanz:
Following https://github.com/romanz/trezor-agent/issues/121


prusnak:
Thanks!

PR #115, opened 2017-06-14
Add 'requests' dependency

romanz:
It's being used by several modules from this package:
```
$ git grep "import requests"
trezorctl:            import requests
trezorctl:            import requests
trezorlib/transport_bridge.py:import requests
trezorlib/tx_api.py:import requests
```


setup.py
```diff
@@ -1,7 +1,13 @@
 #!/usr/bin/env python
 from setuptools import setup
 
-install_requires = ['ecdsa>=0.9', 'protobuf>=2.6.1', 'mnemonic>=0.17', 'setuptools>=19.0']
+install_requires = [
+    'ecdsa>=0.9',
+    'protobuf>=2.6.1',
+    'mnemonic>=0.17',
+    'setuptools>=19.0',
+    'requests>=2.9.1',
```
romanz: > 2.9.1

It's the version I am using now - not sure if it's OK to specify it here.

prusnak: Let's use 2.4.0

romanz: Sounds good, re-pushed.

prusnak:
Thanks!

PR #118, opened 2017-06-21
Allow testing recovery with an initialized device

romanz:
This would allow safe mnemonic validation by using a dry-run flag (following https://github.com/trezor/trezor-mcu/issues/59).

**NOTE**: this PR requires re-generating protobuf-related code after trezor/trezor-common#33 is merged.


jhoenicke:
It doesn't work, because it still complains that the device is already initialized when dry_run is set.
Line 834 in client.py.

romanz:
Oops... my bad.
Will fix later today.

prusnak:
Thanks!

PR #125, opened 2017-08-09
Readme: Click is necessary for latest master

runn1ng:


saleemrashid:
@runn1ng This should be installed by `pip` anyway

prusnak:
Yeah, like Saleem says.

prusnak:
Version published to PyPI (0.7.x) does not require click.

Master requires click, but it is also listed as a dependency, thus 

```
pip install git+https://github.com/trezor/python-trezor@master
```

Should install click as well.

runn1ng:
Oh you are right. OK

PR #126, opened 2017-08-12
Updated unit tests for BitcoinCash

jhoenicke:
Use mainnet, since testnet is not in firmware by default.
Add tests for GetAddress.
Updated txcache.


saleemrashid:
@jhoenicke Unit tests require `DEBUG_LINK=1` to be useful, maybe we should enable the testnet for `DEBUG_LINK=1` or add an extra option like `DEBUG_COINS=1`?

prusnak:
I like the idea.

@jhoenicke If we decided to purse the idea to have extra testnet coins in firmware when `DEBUG_LINK=1` is set, do you think it would be beneficial to not remove Bcash Testnet tests, but keep both? Or Bcash Testnet tests are not that important now that we have mainntest unit tests?

jhoenicke:
I ported all tests  to mainnet, with only miner differences.  I think the testnet tests are no longer necessary.  
I reused two of your wallet tests  :), one with change and one without,  instead of the 2 input, 2 output test, though.

prusnak:
OK, merging then. Thanks!

Issue #129, opened 2017-08-17
ethereum_get_address different from MyEtherWallet


ulope:
(I'm not sure if this is the correct repo to report this but since I'm seeing this issue from the python client it seemed like a good starting point.)

The ethereum address returned by `trezorctl ethereum_get_address` differs from the addresses MyEtherWallet shows when using the Trezor. I assume different methods to derive the key (and thereby address) are being used. It would be nice if there would be a way to get the same addresses MEW uses from the python client.

prusnak:
How exactly do you use `trezorctl ethereum_get_address`? Do you specify correct path?

ulope:
Thats the exact command I'm using. I'm not manually specifying a path.

prusnak:
You need to use the following to generate first address of first account:

```
trezorctl ethereum_get_address -n "m/44'/60'/0'/0/0"
```

ulope:
Good to know. Maybe that should be mentioned in the commands help.

PR #130, opened 2017-08-19
Make -n/--address parameter required.

jhoenicke:
If you really want to have the master public key, you can still just
give the empty string.
* Changed -address to --address everywhere.
* Added help for this parameter.
* Added required flag.


jhoenicke:
Fixes #129.

prusnak:
Missing dash was a typo, of course. Thanks!

PR #133, opened 2017-09-03
fix v1 protocol

mruddy:
small patch to get things working again.
test with: trezorctl get_features


saleemrashid:
@mruddy Have you tested that on both Python versions? I suspect that will break whichever one you are not using

mruddy:
@saleemrashid Yes, I got it working on python2 at least since when running with python3, there were further problems.

prusnak:
I think the correct fix is '`b\x00'`

mruddy:
@prusnak updated, thanks

mruddy:
btw, the further problem that i hit with python3 was due to the hid package:
```
hid/__init__.py, line 79
print 'opening with path'

SyntaxError: Missing parentheses in call to 'print'
```

prusnak:
@jpochyla will look at it tomorrow as he is the one who broke it! :-)

PR #136, opened 2017-09-13
client.py: matrix recovery info: mention backspace

bithobbes:
It is not obvious that it is possible to go back by backspace. Knowing this makes data entry much more comfortable.


prusnak:
Thank you!

PR #138, opened 2017-10-12
Expect confirm button for unusual change address

jhoenicke:
Changed two tests to different (existing) txs with correct change
address, changed the other tests to expect an additional button
confirmation.

See trezor/trezor-mcu#227


saleemrashid:
@prusnak Firmware pull request has been merged, this should probably also be merged (but I have not tested it)

prusnak:
Ay ay. This somehow slipped between my fingers and I have not noticed this earlier.

PR #139, opened 2017-10-23
v2 protocol: bytearray for python2

mcudev:
Tested with `./trezorctl ping test` in bootloader and firmware modes of TREZORv2 using python2


prusnak:
Actually hid_version means something different. It's not about v2 protocol. It's about the change when we decided to remove report_numbers from HID descriptor in the past and use messages without report ID. These messages are sent via byte sequence starting with an extra 0.

mcudev:
ok, yeah i saw the `is_trezor1` and read right past that. still need the `bytearray` part for working with the v2 in non-bootloader mode.

prusnak:
Closed via 54426761c62ac961c7742d51e95f0541b3a48edb

mcudev:
thanks!

PR #141, opened 2017-10-31
Updated segwit_native tests to bech32 and python3

jhoenicke:
Updated addresses from base58 to bech32.

The `map(lambda..., range(1,4))` doesn't work correctly under python3.


prusnak:
Thanks!

PR #148, opened 2017-11-09
Adding note: trezorctl is installed on debian / tails

jonathancross:


prusnak:
The wording is not exactly correct. It is packaged for Debian Stretch, but not in the default install. However, it is in default install for Tails.

jonathancross:
Thanks, fixed now.

prusnak:
Splendid!

PR #155, opened 2017-11-15
Expect confirm button for unusual change address

jhoenicke:
Changed two tests to different (existing) txs with correct change
address, changed the other tests to expect an additional button
confirmation.


teknico:
These changes have already been merged to `master` as part of #138: this is the exact same PR.

jhoenicke:
You're right.

PR #161, opened 2017-11-24
USAGE: Verbose args & native Bech32 segwit

jonathancross:
Seemed that for the target audience (new users) it makes more sense to use the verbose argument form in these examples.  I tested each example to make sure it works. :-)


prusnak:
Thank you!

PR #162, opened 2017-11-25
Parse json floats as string

jhoenicke:
With python-2.7 the float values are sometimes rounded to unacceptable levels, e.g. stripping the last two digits for values over 100k BTC.
This change parses floats as strings to avoid any floating point problems.

Refactored get_url out of fetch_json to make it easier to add new tx_api with a different url scheme.

Note: This will also quote values in the cache.  I consider this a feature, not a bug.  I also tried using `parse_str=Decimal` but then dump doesn't work (known problem with json library).

### DIFF
--- a/trezorlib/tx_api.py
+++ b/trezorlib/tx_api.py
@@ -32,19 +32,23 @@ class TxApi(object):
         self.network = network
         self.url = url
 
+    def get_url(self, resource, resourceid):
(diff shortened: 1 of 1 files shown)


prusnak:
:+1:

saleemrashid:
> but then dump doesn't work (known problem with json library).

Alternatively we could use `simplejson` instead which natively supports `Decimal` (externally maintained version of standard library `json`)

PR #163, opened 2017-11-25
pin: fail if PIN is not numeric

romanz:
When NUM LOCK is disabled, `getpass.getpass()` returns non-numeric string when keypad is used.
I'd prefer `callback_PinMatrixRequest()` to fail early, before sending the invalid PIN to the device (and causing it to increment the failure counter).


prusnak:
Fixed differently in ae663ffe0c2ca11905069932e599cfaca262c064.

(used `isdigit`, because with TREZOR model T, it will be possible to use 0 in PIN)

PR #164, opened 2017-12-01
Better handling of user input for --coin in sign_tx

jonathancross:
* Removes ugly stacktrace for simple user input errors.
* Provides list of valid coin types.
* Both old and new versions have same shell return value: `1`

Example:
```
trezorctl sign_tx --coin btc
Coin "btc" is not recognized.
Supported coin types: Litecoin, Bcash, Bitcoin, Zcash, Dash, Testnet
```

Without this change:
```
trezorctl sign_tx --coin btc
Traceback (most recent call last):
  File "/usr/local/bin/trezorctl", line 819, in <module>
    cli()
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/decorators.py", line 27, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/usr/local/bin/trezorctl", line 453, in sign_tx
    raise ValueError('Coin "%s" is not supported' % coin)
ValueError: Coin "btc" is not supported
```


trezorctl
```diff
@@ -447,10 +447,13 @@ def get_public_node(connect, coin, address, curve, show_display):
 @click.pass_obj
 def sign_tx(connect, coin):
     client = connect()
-    try:
+    if coin in coins_txapi:
         txapi = coins_txapi[coin]
-    except:
-        raise ValueError('Coin "%s" is not supported' % coin)
+    else:
+        print('Coin "%s" is not recognized.' % coin)
```
saleemrashid: Use `file=sys.stderr`

jonathancross: If we want to maintain python 2.x compatibility, I'd have to also add `from __future__ import print_function` at the beginning of the file as [demonstrated here](https://stackoverflow.com/a/14981125/1117929).
I found that `sys.stderr.write()` actually works perfectly on 2.x and 3.x -- seems a bit cleaner as there is no changes to imports / print function, but I'm a python noob, so please advise  :-)

saleemrashid: I'm fine with whichever is consistent with the rest of the code

jonathancross: Okay, went for the `sys.stderr.write()` option then  :-)

prusnak:
Thanks!

prusnak:
I will change it in the upcoming commit, but since we are using `click` we can also use `click.echo` instead: http://click.pocoo.org/5/utils/#printing-to-stdout

jonathancross:
Thanks @prusnak !

PR #165, opened 2017-12-01
Transaction signing example and explanation

jonathancross:
For a while now, I've wanted to provide an actual transaction signing example for users.

Not 100% sure this is the best location, but probably a good start.
Maybe better as a separate file and / or added to https://doc.satoshilabs.com/trezor-apps/ ?
I have screenshots as well which would make for a more friendly version on that site.

Feedback is welcome!


prusnak:
Let's create a `/docs` directory, move USAGE.rst and images there.

jonathancross:
@prusnak Would you like that as part of this PR or as a followup after it is merged?

prusnak:
This PR would be better.

.gitignore
```diff
@@ -4,7 +4,6 @@ python_trezor.egg-info/
 trezor.egg-info/
 *.pyc
 *.bin
-*.png
```
jonathancross: Allow png files again now that we are using these for screenshots.

jonathancross:
Okay, now updated.  :-)

prusnak:
:+1:

jonathancross:
Cheers!

PR #167, opened 2017-12-06
[Stellar] [WIP] Add Stellar support to trezorctl

zulucrypto:
Please see the trezor-common PR for more details and notes on testing: https://github.com/trezor/trezor-mcu/pull/259


trezorlib/stellar.py
```diff
@@ -0,0 +1,52 @@
...
+    else:
+        index = int(account) - 1
+
+    if index < 0 or index > max_account_index:
+        raise ValueError("Error: invalid account number (must be between 1 and " + str(max_account_index) + ")")
+
+    return index
+
+
+def get_public_key_from_bytes(pk_bytes):
```
saleemrashid: Wouldn't `address_from_public_key` be a better name?

zulucrypto: Agreed, renamed to `address_from_public_key`

zulucrypto:
@tsusanka - I've added two tests to this PR. When you get a chance, could you give them a quick look and make sure I'm on the right track?

If they look good, I'll:

 * Write more tests for the remaining operations and commands
 * Add some documentation for using the commands I added to python-trezor
 * Clean up the other PRs (the trezor-mcu one doesn't look right at all)

Thanks for the pointer to the eth test, it was very helpful!

tsusanka:
This looks great! That's indeed exactly what we need. Do the tests pass against your implementation? There might be some nit-picks, but you're definitely going the right direction.

I'm also cc-ing here @matejcik who is somewhat responsible for the python-trezor repository, so he might be able to assist for some python-trezor specific stuff.

zulucrypto:
@tsusanka Yes, the Stellar tests are passing against my (unpushed) implementation of trezor-mcu. However, I'm seeing the test suite hang on `trezorlib/tests/device_tests/test_msg_resetdevice_t2.py`. The device is showing a dialog prompting me to write down one of the seed words. Let me know if you or @matejcik have any ideas.

I'm running the test suite with: `python -m pytest --pyarg trezorlib.tests.device_tests`

trezorlib/client.py
```diff
@@ -1100,6 +1101,172 @@ def self_test(self):
 
         return self.call(proto.SelfTest(payload=b'\x00\xFF\x55\xAA\x66\x99\x33\xCCABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!\x00\xFF\x55\xAA\x66\x99\x33\xCC'))
 
+    @expect(proto.StellarPublicKey)
+    def stellar_get_public_key(self, address_n):
+        return self.call(proto.StellarGetPublicKey(address_n=address_n))
+
+    def stellar_sign_transaction(self, txEnvelope, address_n, networkPassphrase=None):
```
matejcik: please use Python naming convention - no camelCase, `tx_envelope` instead of txEnvelope etc

trezorlib/client.py
```diff
@@ -1100,6 +1101,172 @@ def self_test(self):
...
+            sequence_number=parsed["sequence_number"],
+            timebounds_start=parsed["timebounds_start"],
+            timebounds_end=parsed["timebounds_end"],
+            memo_type=parsed["memo_type"],
+            memo_text=parsed["memo_text"],
+            memo_id=parsed["memo_id"],
+            memo_hash=parsed["memo_hash"],
+            num_operations=parsed["num_operations"]
+        ))
+        if resp.__class__.__name__ != "StellarTxOpRequest":
```
matejcik: why not `isinstance` ?

trezorlib/client.py
```diff
@@ -1100,6 +1101,172 @@ def self_test(self):
...
+            raise CallException("Unexpected response to transaction")
+
+        for opIdx in range(0, parsed["num_operations"]):
```
matejcik: please drop the zero, `range(parsed[...])` is good enough

also `opIdx` is camelCase so please rename that

matejcik: no, actually, the range doesn't seem to make sense here? see below

trezorctl
```diff
@@ -27,6 +27,7 @@ import functools
 import json
 import os
 import sys
+import trezorlib.stellar as stellar
```
matejcik: please move this to the other trezorlib imports
also: `from trezorlib import stellar`

trezorlib/client.py
```diff
@@ -37,6 +37,7 @@
 from .coins import coins_slip44
 from .debuglink import DebugLink
 from .protobuf import MessageType
+from . import stellar as stellar
```
matejcik: drop the `as stellar` please, it's superfluous.
also please move it together with the other `from . import` statements

trezorlib/client.py
```diff
@@ -1100,6 +1101,172 @@ def self_test(self):
...
+        ))
+        if resp.__class__.__name__ != "StellarTxOpRequest":
+            raise CallException("Unexpected response to transaction")
+
+        for opIdx in range(0, parsed["num_operations"]):
+            op = parsed["operations"][opIdx]
+            resp = None
+
+            # Create account
+            if op["type"] == 0:
```
matejcik: this whole series of checks should be its own function(s) in `stellar.py`

also maybe use constants for the types? I imagine something like
```python
OP_CREATE_ACCOUNT = 0
OP_PAYMENT = 1
# ...
def make_operation(op):
    if op["type"] == OP_CREATE_ACCOUNT:
       return proto.StellarCreateAccountOp(...)
    if op["type"] == OP_PAYMENT:
       return proto.StellarPaymentOp(...)
    # ....
```
then in client.py just:
```python
for op in parsed["operations"]:
   resp = self.call(stellar.make_operation(op))
   if isinstance(resp, StellarSignedTx):
        return resp
```

trezorlib/client.py
```diff
@@ -1100,6 +1101,172 @@ def self_test(self):
...
+            # Merge Account
+            if op["type"] == 11:
+                resp = self.call(proto.StellarBumpSequenceOp(
+                    source_account=op["source_account"],
+                    bump_to=op["bump_to"]
+                ))
+
+            # Exit if the response was a StellarSignedTx
+            if resp.__class__.__name__ == "StellarSignedTx":
+                return resp
```
matejcik: this seems that you simply drop every `resp` that isn't a `StellarSignedTx`? what are these?

trezorlib/client.py
```diff
@@ -1100,6 +1101,172 @@ def self_test(self):
...
+        raise CallException("Reached end of operations without a signature")
+
+    @expect(proto.StellarMessageSignature)
+    def stellar_sign_message(self, address_n, message):
+        return self.call(proto.StellarSignMessage(address_n=address_n, message=message))
+
+    def stellar_verify_message(self, pubkey_bytes, signature, message):
+        resp = self.call(proto.StellarVerifyMessage(public_key=pubkey_bytes, message=message, signature=signature))
+
+        if isinstance(resp, proto.Success):
```
matejcik: nooooo, just `return isinstance(resp, proto.Success)` please .)
i know this is how all the other `verify_message` calls look, it's wrong in them too and I'll be changing it

trezorlib/stellar.py
```diff
@@ -0,0 +1,290 @@
...
+            raise ValueError("Ending timebound out of range (must be between 0 and " + max_timebound)
+
+    # memo type determines what optional fields are set
+    parsed["memo_type"] = unpacker.unpack_uint()
+    parsed["memo_text"] = None
+    parsed["memo_id"] = None
+    parsed["memo_hash"] = None
+
+    # text
+    if parsed["memo_type"] == 1:
```
matejcik: maybe consider named constants here too?

trezorlib/stellar.py
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    has_source_account = unpacker.unpack_bool()
+    if has_source_account:
+        op["source_account"] = unpacker.unpack_fopaque(32)
+
+    op["type"] = unpacker.unpack_uint()
+
+    # see: https://github.com/stellar/stellar-core/blob/master/src/xdr/Stellar-transaction.x#L16
+    if op["type"] == 0:
```
matejcik: ...and also here

matejcik:
@zulucrypto as you can see, I did a review of your code :) please look at it and make changes where appropriate.

also, your commit is removing a bunch of fields from `DebugLink` messages. I don't know why that is happening - were you perhaps using an older version of `trezor-common` to generate them? I suspect that _might_ be the cause of your hanging test.

zulucrypto:
@matejcik - Thanks for the detailed comments, as you can tell I'm not very familiar with python and they're very helpful!

> were you perhaps using an older version of trezor-common to generate them? 

I think you're right, I'll re-sync master to my trezor-common branch and try again.

trezorlib/client.py
```diff
@@ -1100,6 +1101,172 @@ def self_test(self):
...
+            # Create account
+            if op["type"] == 0:
+                resp = self.call(proto.StellarCreateAccountOp(
+                    source_account=op["source_account"],
+                    new_account=op["new_account"],
+                    starting_balance=op["starting_balance"]
+                ))
+            # Payment
+            if op["type"] == 1:
+                asset = types.StellarAssetType(type=op["asset"]["type"], code=op["asset"]["code"],
```
matejcik: just noticed: `types` doesn't exist, you probably meant `proto.StellarAssetType`

matejcik:
revisiting the code
it would be better if your XDR decoding functions returned protobuf messages directly, instead of roundtripping through dicts.

e.g., `_xdr_read_asset` could directly return a `StellarAssetType`

trezorlib/client.py
zulucrypto: The signing loop for Stellar looks like this:

1. Start with a Transaction that contains a Header and N Operations
2. Send the Header to the device
3. Receive a `StellarTxOpRequest` message
4. Send Operation 1 (protobuf message varies depending on operation type)
5. Receive `StellarTxOpRequest` (device expects another operation) OR `StellarSignedTx` (that was the last Operation in the Transaction)
a. If a `StellarSignedTx` was received, return it since this is what we're after
b. Otherwise, go to 4 and send Operation 2

Since the python code and the device should both agree on the number of operations I could remove the check for whether a `StellarSignedTx` came back and instead move this outside the loop and return it after all operations were sent, if you think that would read better?

zulucrypto:
Hi @matejcik - I believe I've addressed your concerns from the first round of review. Good call on returning protobuf messages directly, it cleaned things up a lot!

I'm currently trying to fix an issue where the firmware is too large to fit on the device, but once I get that taken care of (or the emulator working) I'll continue writing tests.

Feel free to look things over again now or I'll ping you when I've got the tests done. Thanks again for the review!

trezorlib/client.py
```diff
@@ -1100,6 +1101,46 @@ def self_test(self):
...
+
+        # Will return a StellarTxOpRequest
+        resp = self.call(parsed["tx"])
+        if not isinstance(resp, proto.StellarTxOpRequest):
+            raise CallException("Unexpected response to transaction")
+
+        # Send each operation to the device.
+        # The response to the last message should be a StellarSignedTx
+        for op in parsed["operations"]:
+            resp = self.call(op)
```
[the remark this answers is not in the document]
matejcik: There should also be a comment preceding this whole block, explaining what is going on.
this is a good candidate for that comment: https://github.com/trezor/python-trezor/pull/167#discussion_r182858940 :)

matejcik:
this looks very good now.

Thanks for the explanation of the signing flow. What I disliked originally (and still do to some extent) is that it was unclear what the expected behavior is. I think that I now came up with something that is clearer in that regard.

Other than that, I have only some minor improvements. The main thing is probably marking the tests properly; `xfail` means that the test is executed, but it's expected to fail. For your testing, you can run `pytest --runxfail` and the tests will behave as normal. For us, we can merge the tests and they will not break the test suite before you merge your changes.

anyway, thanks and keep up the good work :)

zulucrypto:
> so how about we turn the loop the other way around and let the device drive it

Thanks for the example code, I've used what you suggested and have added a comment that explains it in more detail.

> are you pushing Stellar support to both T1 and T2?

I would like to, but haven't tested it on the T2 yet or looked into whether I need to compile a separate firmware.

> Your tests don't seem to use anything else right now, but if they do, that should be imported separately (from binascii import unhexlify is a common one)

I'm using some functions from base64 in `test_msg_stellar_sign_transaction.py`, is this import correct?

`from base64 import b64decode, b64encode`

Although I'm still trying to figure out a workaround for the firmware size limit, the code is ready for another look. I appreciate your code examples and the explanation about `xfail` and how to run the tests!

trezorlib/stellar.py
```diff
@@ -0,0 +1,323 @@
...
+    # id (64-bit uint)
+    if tx.memo_type == MEMO_TYPE_ID:
+        tx.memo_id = unpacker.unpack_uhyper()
+    # hash / return are the same structure (32 bytes representing a hash)
+    if tx.memo_type == MEMO_TYPE_HASH or tx.memo_type == MEMO_TYPE_RETURN:
+        tx.memo_hash = unpacker.unpack_fopaque(32)
+
+    tx.num_operations = unpacker.unpack_uint()
+
+    for i in range(tx.num_operations):
```
matejcik: please move the `operations = []` from above here, so that it's clearer where it's coming from.
(it's not used before here, right?)

trezorlib/stellar.py
```diff
@@ -0,0 +1,323 @@
...
+
+def _parse_operation_bytes(unpacker):
+    """Returns a protobuf message representing the next operation as read from
+    the byte stream in unpacker
+    """
+
+    # Check for and parse optional source account field
+    source_account = None
+    has_source_account = unpacker.unpack_bool()
+    if has_source_account:
```
matejcik: better: `if unpacker.unpack_bool():` as you do in `OP_SET_OPTIONS`

trezorlib/stellar.py
```diff
@@ -0,0 +1,323 @@
...
+
+    Ported from Java implementation at: http://introcs.cs.princeton.edu/java/61data/CRC16CCITT.java.html
+
+    Initial value changed to 0x0000 to match Stellar configuration.
+    """
+    crc = 0x0000
+    polynomial = 0x1021
+
+    for byte in bytes:
+        for i in range(0, 8):
```
matejcik: `for i in range(8)`

trezorlib/stellar.py
```diff
@@ -0,0 +1,323 @@
...
+            bit = ((byte >> (7 - i) & 1) == 1)
+            c15 = ((crc >> 15 & 1) == 1)
+            crc <<= 1
+            if c15 ^ bit:
+                crc ^= polynomial
+
+    return crc & 0xffff
```
matejcik: please add a newline at end

matejcik:
hm, i tried to comment via e-mail and that apparently didn't go through. oh well.
anyway.
Yes, your `from base64 import` is ok, same as would be `import base64` and then `base64.b64encode`. It's up to you.

(what I wrote about `binascii.hexlify` is something slightly different: in many of our existing tests, `hexlify` is imported with `from .common import *`. To avoid having to change the test code, one can add `from binascii import hexlify`. But it is as valid as `import binascii` and then using `binascii.hexlify` in the test.)

Anyway, the code looks good to me now, basically OK to merge. Do you want to do that or do you want to wait for the `trezor-mcu` PR?

matejcik:
oh, except of course you need to fix the conflicts.
and also there are probably `flake8` complaints (wrong number of whitespace) etc.

prusnak:
>  Do you want to do that or do you want to wait for the trezor-mcu PR?

I am OK with merging this code now, i.e. don't wait for trezor-mcu merge.

zulucrypto:
@matejcik - I've done some more cleanup to take care of your improvements, flake8 style issues, and to resolve the merge conflicts.

> Anyway, the code looks good to me now, basically OK to merge. Do you want to do that or do you want to wait for the trezor-mcu PR?

I'd like to add some more tests and documentation, although I'm blocked on the tests due to the firmware size limit.

@tsusanka - Do you have any suggestions for working around the firmware size limit or a way I can run the tests? Also, are you still waiting on anything from me to do more review on the `trezor-mcu` PR?

prusnak:
@zulucrypto as per comment: https://github.com/trezor/trezor-mcu/pull/259#issuecomment-385957947 I resolved the firmware size limit issue, so you can hopefully test again

prusnak:
JFYI I just merged Stellar changes into trezor-mcu via https://github.com/trezor/trezor-mcu/pull/259

PR #174, opened 2017-12-17
Refactor pb2py

saleemrashid:
This is a very large and opiniated pull request so it is unlikely that you will want to merge it entirely but you may want to cherry-pick parts of it.

* Use `importlib.import_module` instead of `__import__`
* Use `argparse.FileType`
* Generate messages and fields in the same order as the source files
* Removed use of `google.protobuf.internal`
* Removed use of `from __future__ import absolute_import`

Note that I haven't run `tools/build_protobuf` in this pull request as it would make the changes harder to review


prusnak:
@slush0 @jpochyla can you please look at the change? In general, I like it. But don't see the big picture how you plan to use this exact code in T2, so review is welcome, before I pull this in.

saleemrashid:
@alepop Could you try this pull request on OS X?

alepop:
@saleemrashid `build_protobuf` works fine now on OSX

matejcik:
I have merged the changes (there was significant divergence) and did some minor modifications. @alepop, could you check that the version currently in master still works on OS X? Thanks.

alepop:
@matejcik everything works fine 👍

PR #180, opened 2017-12-19
Fix OutputScriptType usage in trezorctl

saleemrashid:


saleemrashid:
Just putting another fix in here as well

PR #181, opened 2017-12-19
Remove default curve

saleemrashid:
Bitten by this when testing Decred.


prusnak:
@romanz Could you please confirm this is OK for your usage in https://github.com/romanz/trezor-agent ?

saleemrashid:
AFAICT it should be fine because https://github.com/romanz/trezor-agent doesn't reference `trezorlib.client.DEFAULT_CURVE`. This doesn't actually change behaviour for coins that use `secp256k1` (the device will select `secp256k1`). This only changes behaviour for coins such as Decred, which uses `secp256k1_decred`.

prusnak:
I checked that trezor-agent indeed does not call the patched 3 methods with unset ecdsa_curve_name parameter.

saleemrashid:
@prusnak Even if it did, it would not be a problem as the device would default to `secp256k1` (which is what `DEFAULT_CURVE` was set to)

romanz:
Looks good, thanks for the fix :)

Issue #123, opened 2017-07-16
get_transport must succeed in order to access command usage


saleemrashid:
This prevents command usage being accessed if no TREZOR is connected or the TREZOR is not accessible.

## Steps to reproduce

1. Disconnect all TREZOR devices
2. `trezorctl get_address --help`

## Expected result

`get_address` usage information

## Actual result

```
Exception:  Device not found
```

prusnak:
I think this would be best fixed as a major rewrite of transports (i.e. transport opens connection only when needed = before send and not when it is instantiated).

I tried to fix trezorctl meanwhile until we'll do this, but I did not find any solution that is not very hackish/ugly. :-/

saleemrashid:
@prusnak I think that RAII makes sense for the transports

prusnak:
The current architecture does not really make sense, frankly.

Some time ago, Jan tried to refactor it (branch https://github.com/trezor/python-trezor/tree/transport-connection), I even pushed it further (branch https://github.com/trezor/python-trezor/tree/protocol-device). But the development stalled because of other work to do.

In the future, I'd like to do something similar to `protocol-device` branch, because all you should care about is some kind of high-level USBDevice/BridgeDevice object, not some stuff transports/protocols/etc. underneath.

saleemrashid:
@prusnak Agreed. Devices as non-RAII, but transports as RAII, makes sense.

jpochyla:
Should be fixed in master, closing?

saleemrashid:
@jpochyla ACK

PR #191, opened 2017-12-26
Add pass_client decorator to trezorctl

saleemrashid:
This replaces the `connect` function for instantiating the client with a decorator.

/cc @jpochyla


matejcik:
The code has drifted too far in the meantime :/ I thought that the main cause of conflict was #226, where I rewrote the code for path matching. But there's actually a lot more conflicts.
Would you care to fix them?

matejcik:
...actually, reading it again, I was under the impression that the reason for having `connect()` work this way is so that the connection is deferred until after _some code_ runs. Most likely instance of "some code" is the firmware download in `firmware_update()`, and possibly also the checks in `set_homescreen()`.
Otherwise you could just assign the client directly to `ctx.obj` and be done with it, no need for a custom decorator.
But your PR loses this anyway, so I'm not sure if this is a good way to do things.

Thoughts?
Maybe leave `cli()` as is, have the decorator only call `connect()`, and use it everywhere except the firmware and homescreen functions?
(I'm not convinced that it's worth it though...)

saleemrashid:
>  Otherwise you could just assign the client directly to `ctx.obj` and be done with it, no need for a custom decorator.

I believed the issue was that this would attempt to connect to the device, even for `--help` (see #123)

matejcik:
hmm, yes, it makes sense in that case.

please fix the conflicts and i'll merge this.

matejcik:
i'm going to close this PR. please reopen if you ever have the time to fix the conflicts

PR #192, opened 2017-12-26
sign_tx: add rbf opt-in enable, add locktime, add tx version, sign based on path

mcudev:
i had this and figured i'd share it. it's simple and scratches some itches.
some other PRs might be going in a different direction, if so, please disregard.


saleemrashid:
@mcudev Could you rebase it on #189 perhaps? I think that PR takes a better approach for script type detection

mcudev:
@saleemrashid done please see the last commit here, thanks

saleemrashid:
@mcudev The input amount is also required for coins that use BIP-143 and other possible future developments, so I would prefer to change that back.

mcudev:
@saleemrashid makes sense, done

prusnak:
Something weird happened here, rebased and merged in 49622077036c92def595800505e4b3940212b724

Please check that's correct.

mcudev:
looks good, thanks!

PR #193, opened 2017-12-28
Allow sign_tx to be less dependent on network stack

brianddk:
Modified how txes are cached so that pulling from network resources only happens when absolutely required

Pros:

* Better functionality in air-gapped tails environment

Cons:

* Minimal testing
* Network failures during signing leave display at "signing" status


matejcik:
I'm not sure how this helps in an airgapped environment -- you still need to download at least one transaction, no?

In any case, this would be better if the `tx_dict` class extended `collections.defaultdict` (see https://docs.python.org/3.6/library/collections.html#collections.defaultdict ) and if you checked for presence of `tx_api` right in the constructor.

I'll also look into whether it's possible to reset the device if signing fails.

brianddk:
Yes, extending `collections.defaultdict` would be better.  Thx.

When I walked through this in Bitcoin Testnet TXNs I found that the default (2 input, 2 output) TXN generally went through the `if not res.details.tx_hash:` path.  The current production code is fetching TXNs when the `None` is requested.  Possibly a much simpler fix would be to put guard-code around that case, I just tried for a more generalised solution.

brianddk:
To clarify, This PR will not fetch TXN in the `if not res.details.tx_hash:` code path.  The current master branch always fetches TXNs even in the `if not res.details.tx_hash:` case.

matejcik:
It turns out that for signing, you need exactly those inputs that are not SegWit enabled.
So it's possible to determine before starting the signing process. In the usual case these days, all inputs are going to be segwit, so there's no downloading necessary. If you're spending from old transactions though, you will still need the full transaction data.

I have refactored the relevant code and it now skips downloading for inputs it doesn't need. If that is all of them, the transaction can be signed offline.

PR #194, opened 2018-01-04
client: expected field check fix

tsusanka:
This fixes an error in https://github.com/trezor/python-trezor/commit/a27217811b8edb9460abede1665c9fecc824907d where the expected fields were not checked


prusnak:
:+1: 

refactored a bit, so also the print statement is skipped (https://github.com/trezor/python-trezor/commit/f587135b8d9f1e5a9a4a9fa0b3142d5b7b8b11e2)

PR #197, opened 2018-01-06
Add Lisk support to trezorctl

alepop:
Messages: trezor/trezor-common#68

Progress: 
- [x] Lisk get address
- [x] Lisk get public key
- [x] Lisk sign message
- [x] Lisk verify message
- [x] Lisk sign tx
  - [x] send transaction
  - [x] signature ransaction
  - [x] delegate transaction
  - [x] vote transaction
  - [x] multisignature transaction
_`dApp` tx types wil not allowed in the Lisk 1.0 release_
~~dapp tansaction~~
~~dapp In transfer transaction~~
~~dapp out transfer transaction~~


prusnak:
@matejcik I merged in trezor-common changes for Lisk messages in https://github.com/trezor/trezor-common/commit/6a7710c026d748f09aca1d95bc82995c52a60479

The API is VERY similar to Bitcoin API.

Can you please review this PR to python-trezor?

alepop:
@prusnak `skip_t1`added

matejcik:
@prusnak the "generated messages in git" are to support people who don't have (the appropriate version of) `protoc` -- of course you need that to contribute the regenerated messages, but users who checkout the repository or use the sdist from PyPI don't need it. Not to mention Windows where you'd need to run a shell script.
Given that it's a single-command step for contributors, and that you can regenerate messages and bump trezor-common in one commit (so no churn), to me it seems a reasonable tradeoff.

prusnak:
>  to me it seems a reasonable tradeoff

Agreed!

trezorlib/client.py
```diff
@@ -834,6 +834,80 @@ def mosaic_supply_change_to_proto(mosaic_supply_change):
...
+        message = normalize_nfc(message)
+        return self.call(proto.LiskSignMessage(address_n=n, message=message))
+
+    def lisk_verify_message(self, pubkey, signature, message):
+        message = normalize_nfc(message)
+        try:
+            resp = self.call(proto.LiskVerifyMessage(signature=signature, public_key=pubkey, message=message))
+        except CallException as e:
+            resp = e
+        if isinstance(resp, proto.Success):
```
matejcik: please change this to `return isinstance(...)`
(yes, all other verify_message functions have this. that will be changed, let's not put it into new code)

matejcik:
the code looks mostly good.

however what I see as a problem is that several clearly numeric fields are declared as strings?
that doesn't seem right - unless that's intentional (Lisk docs seem to suggest that it isn't), it should be fixed. in trezor-common first and then in the other PRs

trezorlib/client.py
alepop: @matejcik done

prusnak:
Please wait with the merge until we resolve int/string discussion in https://github.com/trezor/trezor-core/pull/90

matejcik:
@alepop when the int conversion goes through, please rebase this PR onto current master, update the `trezor-common` submodule and regenerate the protobuf messages. thanks.

alepop:
@matejcik done.

matejcik:
looks good, thanks

PR #198, opened 2018-01-06
client: Fix expected_responses

saleemrashid:
This is kind of a band-aid solution to make the tests pass after #194 because we cannot tell which fields are expected to be missing. To completely fix this, we will probably need to refactor the code a bit.


prusnak:
Fixed in 0c5eac2f39840c9a48511544a0b6dc7ecbe0e9ff

PR #200, opened 2018-01-11
tests/device: ethereum sign tx with expected checks

tsusanka:
Checks expected responses in ETH signing. I was unable to test this against T1, I'm having some troubles running the tests on T1.


prusnak:
Fixed test in 47cfa178e4b964e92550c2671153d2e54f948caf

For Ethereum we always use ButtonRequestType == SignTx (even when confirming "outputs").

PR #202, opened 2018-01-14
Fixes for BlockCypher API and Dogecoin.

antonkolesnyk:
Either some things have changed a bit recently, or my initial commit for Dogecoin sign_tx support did not cover all the cases.

I've found out that if one of the inputs comes from rather a complex transaction with multiple inputs and outputs, I am getting "Encountered invalid prevhash" error, unless I use BlockCypher's API via httpS instead of http (while simple transactions with one input and two outputs do work fine if we use plain http).

And also BlockCypher may omit lock_time in the JSON it returns, if the value is zero.


prusnak:
:+1:

PR #203, opened 2018-01-14
 Add varying url to broadcast tx, depending on the api

antonkolesnyk:
Every API has different URL for sending the transaction that you get from sign_tx.
Previously, in the code, it would give correct URL only for the Insight API.
Now, it gives the correct url for BlockCypher API (Dogecoin) as well.

It is now possible to also specify the correct url for SmartBit API. Unfortunately, I was unable to find the correct API URL for them, so for SmartBit the behavior will still be the same as it was before, meaning that the URL that gets printed in the end is unlikely the one that can be used to send the transaction.
I found their URL for Bitcoin - https://www.smartbit.com.au/txs/pushtx - but it is not the coin Trezorctl uses it for.
By the way, their API which is used in the Trezorctl's code for Segnet - https://segnet-api.smartbit.com.au/v1/blockchain/ , seem to be dead.

Please let me know if the name I chose for pushtx_url is good, or if I should rename it. My other suggestions for the name were sendtx_url, or something involving "broadcast", or "relay". Also I am not finally sure if the naming style with the underscore is consistent with the style that is used in Trezorctl.

### DIFF
--- a/trezorctl
+++ b/trezorctl
@@ -552,7 +552,7 @@ def sign_tx(connect, coin):
     click.echo(binascii.hexlify(serialized_tx))
     click.echo()
     click.echo('Use the following form to broadcast it to the network:')
-    click.echo(txapi.url.replace('/api/', '/tx/send'))
+    click.echo(txapi.pushtx_url)
 
 
 #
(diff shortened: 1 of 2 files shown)


prusnak:
I like the name `pushtx_url` so I accepted it. :+1:

PR #205, opened 2018-01-23
Add Terracoin support

TheSin-:
Adding Terracoin support for trezor here too.


prusnak:
After https://github.com/trezor/python-trezor/commit/bd43363b1cbc3d966a9b30747a17f1616c00316e merged, coins now use generated definitions from coins.json, so a commit to trezor-common to include the coin definition is enough. Closing.

PR #209, opened 2018-02-01
Add webusb to transports

karel-3d:
V2 protocol with debug link is not tested.


prusnak:
Nice! Will add some nits in another commit. And also we need to test this whether this works in Windows and macOS.

PR #214, opened 2018-02-04
Allow compatibility with Python 2

romanz:
Following https://github.com/romanz/trezor-agent/issues/195


prusnak:
I will merge this but please note that we won't be keeping py2 compatibility for long.

PR #222, opened 2018-02-28
pretty printing protobuf structures

matejcik:
Instead of dumping the one-line `repr` blob, do something similar to what `pprint.pprint` does.

useful for `trezorctl get_features` and for watching verbose output

Special tweaks exist for:
* `address_n` which is an array of small numbers so goes on one line instead of one-per-item
* `bytearray`s dump their contents hex-encoded
* `Features` are now omitted from debug outputs because they're very long and seen on every session again and again


matejcik:
@saleemrashid I looked into extending PrettyPrinter and it seems rather unwieldy; several people recommend not doing that. But if you have a good example that could be useful here, I'm not against converting the code to it.

PR #224, opened 2018-03-02
tests: ethereum sign/verify is skipped

tsusanka:
Skip ethereum message signing and verification. It is not yet implemented and waits for http://github.com/ethereum/EIPs/pull/712

cc @matejcik


prusnak:
:+1:

PR #226, opened 2018-03-02
prefix search for `trezorctl -p`

matejcik:
The new feature:
`trezorctl -p hid` will pick the first HID device, and `trezorctl -p bridge:w` will pick the first WebUSB device on bridge. It's still possible to specify a full path and that will be tried directly before prefix-searching (so it's possible to specify UDP addresses that are not the default one).

There's also a refactor of transports.

While the prefix search for paths is desirable, the refactor is not necessary, I just took the opportunity to clean up a little.

The main thing is removing `device.py` because it's very much useless. The functionality is now moved to a pair of functions in `trezorlib.transport`. This is an API change and will need to be fixed in users... which is why i'm sort of hesitant about it. OTOH, the original API doesn't make sense.

I'll probably add a backwards compatibility wrapper that shouts DeprecationWarnings.

(it's also easy to write backwards compatible code:
```python
try:
    from trezorlib.transport import enumerate_devices
except ImportError:
    from trezorlib.device import TrezorDevice
    enumerate_devices = TrezorDevice.enumerate
```
)

Transports now have their own submodule, which seems cleaner but isn't really necessary either.


matejcik:
posting this now to solicit comments. It's still necessary to update most tools to use the new API, and do the DeprecationWarning wrapper.

prusnak:
@slush0 would you like to have a look?

saleemrashid:
Since you're changing the transports API, could you also update `tools/*.py`? As it is, they don't work with T1 >= 1.7.0 or T2.

matejcik:
@saleemrashid does this solve your problem?

also added the promised `TrezorDevice` deprecation wrapper.

prusnak:
LGTM

PR #227, opened 2018-03-04
Update ethereum_sign_tx to python3

jhoenicke:
Should now work with both python2 and python3.

Currently, it only works when giving `nonce`, `gas-price` and `gas-limit` on the command line, because ethjsonrpc itself is buggy and not python-3 compatible.


prusnak:
Thanks!

PR #228, opened 2018-03-04
Wanchain support in python-trezor

tyrion70:
Added wanchain_sign_tx function to call the new wanchain_sign_tx function in trezor-core. Updated Usage as well, since wanchain requires an extra parameter txtype (-x).


prusnak:
Closing this. Please reopen with new changeset once we have merged the trezor-common change (ideally minimal required change to get this working).

PR #229, opened 2018-03-06
Move to Python 3 only code

matejcik:
This is a pull request to drop vestiges of Python 2 support and clean up the code somewhat in the process.

It also asks Travis to stop testing python 2.7.


prusnak:
LGTM

PR #233, opened 2018-03-11
tests: add test_msg_getecdhsessionkey.py

romanz:
Following https://github.com/trezor/trezor-core/pull/158 - tested on both TREZOR One and TREZOR Model T emulator.


prusnak:
Thank you!

PR #235, opened 2018-03-13
Allow setting auto-lock delay

romanz:
This PR depends on https://github.com/trezor/trezor-common/pull/87 and https://github.com/trezor/trezor-mcu/pull/315.


romanz:
@prusnak Should I add a simple test here?
https://github.com/trezor/python-trezor/blob/master/trezorlib/tests/device_tests/test_msg_applysettings.py

matejcik:
@romanz yes, please add the test

matejcik:
Hmm, that's a rather long delay.
OTOH i suppose the minimum delay is there for a reason. Perhaps the test should also check that you cannot set a lower value.
Then maybe we can mark the delay test somehow so that it can be easily skipped?

prusnak:
I changed the minimum delay to 10 seconds in https://github.com/trezor/trezor-mcu/commit/783f1c0323013244a442778ae9b3cbec4e449a5b

You can rework the test to use the new value and I agree with @matejcik that we can add another test that setting to e.g. 5 seconds does not work (and still at least 10 seconds are used).

romanz:
Sounds good, I've updated and pushed a new commit.

matejcik:
looks mostly good now.
one last thing, given that minimum is 10s, maybe the trezorctl command should take seconds instead of minutes? (or even "60" / "60s" / "1m")

matejcik:
ummm no, let's not add a dependency for this!

I was thinking of something like:
```python
if delay[-1] == "m":
  seconds = 60 * int(delay[:-1])
elif delay[-1] == "s":
  seconds = int(delay[:-1])
else:
  seconds = int(delay)
```

(Sorry for being so nitpicky. But this feature is simply not important enough to warrant a new dependency, however small. I'd also be OK with just switching from minutes to seconds, without any suffix logic whatsoever.)

romanz:
No problem :)
Updated the PR.

matejcik:
perfect, thank you :)

prusnak:
Thank you Roman! Your contributions are very much welcome!

romanz:
Happy to help :)

PR #238, opened 2018-03-16
Refactor code generation

saleemrashid:
Minor changes

* Move `import` generation for messages to separate function
* Use generators instead of lists (`yield "generated code"` rather than `out += ("generated code", )`) 
* Use `logging` module instead of `print`


matejcik:
sure :+1:

PR #242, opened 2018-03-28
trezorctl: update url to releases.json

bithobbes:


prusnak:
I think we can implement the following improvements:

a) use `connect().features.bootloader_mode` to check whether the device is in the bootloader mode first and throw an exception if not

b) use `connect().features.major_version` to determine whether we should put `1` or `2` in the JSON url

prusnak:
Closing as unmerged. This needs to be handled in a more systematic way.

Let's deal with that in https://github.com/trezor/python-trezor/issues/269

PR #248, opened 2018-04-05
use trezor-common as a submodule

matejcik:
- [x] Attempt to build protobuf messages from `trezor-common` as part of `setup.py`. This falls back to in-tree versions if `build_protobuf` fails, e.g. if you don't have `protoc` or are on Windows where shell scripts don't work
- [x] make Travis check that in-tree protobuf messages are really generated from trezor-common
- [x] put a note in README saying how to update the submodule
- [x] install `coins.json` from trezor-common as part of the package
- [x] WIP - use `coins.json` in `coins.py` instead of static dictionaries

also flake8 is better and stricter here


prusnak:
Looks great! ACK 12ea96a

prusnak:
:tada:

PR #252, opened 2018-04-21
lisk: Fix wrong signature in tx tests

alepop:
I found a bug in logic in the `trezor-core` module for lisk signing.
This is a fixed test for right logic


matejcik:
merged, thanks

PR #260, opened 2018-05-11
code refactor - first step (codename "verbosity")

matejcik:
this is a lot of churn, and a bunch of small changes:

* crypto utilities that are only used in tests -- that is, `ed25519raw`, `ed25519cosi` and `ckd_public` have been moved to `trezorlib.tests.support`
  Functions in `trezorlib.tools` are next on the chopping block; what holds them is a single method `load_device_by_xprv` which is the only thing that needs to know how to decode a b58-encoded key. This will probably be moved somewhere (the tools dir? a debug module?)  and we might grow an optional dependency on some sort of actual BTC format library.
* `trezorlib.tools.Hash` is renamed to `btc_hash`, because naming convention. Basically nothing in the library is using this and you shouldn't be using it either.
* `TrezorClient.expand_path` method is changed into a `trezorlib.tools.parse_path` function, with some minor improvements. For instance, you can now use `1h` instead of `1'`, which is friendlier to type.
* `PRIME_DERIVATION_FLAG` no longer exists. The constant is moved to `tools` and renamed to `HARDENED_FLAG`, inline with BIP32 update. ("prime derivation" is difficult to google and has this sort of "you had to be there" vibe to it). You usually shouldn't use it though; If you want to "harden" a number, call `tools.H_()` on it.
* `client.format_protobuf` is moved to `protobuf.format_message`, where it arguably belongs better. It has only existed for one micro version, so hopefully nobody is using it yet.
* `client.AssertionException` is gone, replaced by an actual `AssertionError`
* flake8 is still stricter and disallows star imports. That means that every single test needed to be touched.

-----

The biggest change is related to the way verbosity (`trezorctl -v`) is accomplished.

`VerboseWireMixin` was dropped, along with `TrezorClientVerbose`. I dearly hope that this won't affect anyone; but if it does, feel free to open an issue about it. You're not getting the mixin back in any case (mixins will go away completely in 1.0), but a backwards-compatible "verbose class" can be added.

Instead, messages are logged through Python's `logging` mechanism. This is now called from protocols, but it will be moved to transports, probably even before this PR goes through. (writing this description I realized that BridgeTransport doesn't actually use a protocol and therefore doesn't log anything.)

There's a new `log` module with three features:
* `PrettyProtobufFormatter` class that outputs formatted messages
* `enable_debug_output()` function that will apply this formatter to a `LogHandler` that you pass, or create a new `StreamHandler` for you if you pass nothing.
  Basically, if you want verbose protobuf, call `trezorlib.log.enable_debug_output()` and you get mostly the same thing as if you used `TrezorClientVerbose`.
  This is what `trezorctl -v` does. And what the test suite does too, but this time, only if you run with `-v`. Running tests normally will show nothing.
* `OMITTED_MESSAGES` is a set of protobuf types that will not be printed verbosely. `trezorctl` puts Features there. This might go away.

In addition, the message formatter is significantly improved and should produce more pleasant output: it does smart(er) shortening of byte buffers, lists of simple items, and can automatically omit the list of coins from the Features message.

With the use of `logging`, client's `log` function is thrown away and replaced by printing to stderr where appropriate (that is, in `TextUiMixin`)


prusnak:
I don't like moving ed25519raw, ed25519cosi into
trezorlib.tests.support. They are being used outside tests too (for
example in trezor-core/tools/keyctl). I would suggest location like
trezorlib.crypto.ed25519 or something similar if you want to
refactor it.
OTOH we might want to drop ckd_public completely and use implementation
from pybitcoin or bip32utils module.

matejcik:
How about this?
I would like to split `TrezorClient` class into a core that performs communication with Trezor, and a bunch of function-specific modules. E.g., instead of  `TrezorClient.nem_sign_tx(self, ...)`, I'd do `trezorlib.nem.sign_tx(client, ...)`.

Then i would keep `ed25519raw` as-is, and fold `ed25519cosi` into `cosi.py` along with `cosi.commit(client, n, data)` and `cosi.sign(client, n, ...)`

prusnak:
LGTM

matejcik:
@prusnak care to look at db92b13 and comment?
if it looks OK to you, I'd update trezor-core's keyctl to use it

trezorlib/cosi.py
```diff
@@ -26,13 +26,24 @@ def get_nonce(sk, data, ctr):
...
+    return ed25519raw.encodeint(S)
+
+
 def self_test(digest):
 
     def to_hex(by):
         return binascii.hexlify(by).decode()
 
     N = 3
-    keyset = [0, 2]
```
prusnak: I wanted to check whether cosi works when only 2 of 3 keys are used. Your change tests only 3 out of 3 scenario, which we don't use in production.

Ideally we should test all combinations of keys, that is: `[1], [2], [3], [1,2], [1,3], [2,3], [1,2,3]`.

prusnak:
Wonderful! utACK 8410151

PR #262, opened 2018-05-16
stellar: xdr parse move to trezorctl; client works with messages

tsusanka:
Is this okay? I think it'll make writing tests easier, since it is clear what is inside the transaction


matejcik:
now that you're passing `tx` and `operations` from the outside, it's fragile to rely on `num_operations` in the `tx` structure. At least that field should be checked and overwritten in the `stellar_sign_tx` method -- probably simply `tx.num_operations = len(operations)`

i would consider passing `address_n` (and possibly `network_passphrase`) separately too, reason being API consistency; both of these fields _don't_ come from the decode method, they're user-supplied, and this would be the only (or one of few) method that basically receives arguments as a prefilled protobuf, as opposed to actual arguments.

in any case, the correction of `network_passphrase` should happen at the same place where `tx` is prefilled -- either we allow an optional argument `network_passphrase` that gets corrected to the "default network passphrase" in the client, or we consume a prefilled `tx` and if the `network_passphrase` is missing, then that's probably deliberate.
I'd probably check this in two places: take an argument in `stellar_sign_tx` and correct it if none, *and* set a default for the `trezorctl` argument

matejcik:
also `expand_path_or_default` is horrible and i should have caught it in the original review o_O but i'll fix that

(for the record `client.expand_path` is obsolete in favor of `tools.parse_path`)

trezorlib/tests/device_tests/test_msg_stellar_sign_transaction.py
```diff
@@ -37,15 +39,35 @@ def get_address_n(self):
     def test_sign_tx_bump_sequence_op(self):
         self.setup_mnemonic_nopin_nopassphrase()
 
-        xdr = b64decode("AAAAABXWSL/k028ZbPtXNf/YylTNS4Iz90PyJEnefPMBzbRpAAAAZAAAAAEAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAt//////////wAAAAAAAAAA")
+        op = proto.StellarBumpSequenceOp()
+        op.bump_to = 9223372036854775807
```
prusnak: Can we write this as 0x7fffffffffffffff ? (It's easier to see that this value is valid).

tsusanka: Definitely, I haven't realized this.

trezorlib/tests/device_tests/test_msg_stellar_sign_transaction.py
```diff
@@ -37,15 +39,35 @@ def get_address_n(self):
...
+
+        response = self.client.stellar_sign_transaction(tx, operations)
         assert b64encode(response.signature) == b'gjoPRj4sW5o7NAXzYOqPK0uxfPbeKb4Qw48LJiCH/XUZ6YVCiZogePC0Z5ISUlozMh6YO6HoYtuLPbm7jq+eCA=='
+
+    def _create_msg(self, operations_count) -> proto.StellarSignTx:
+        tx = proto.StellarSignTx()
+        tx.protocol_version = 1
+        tx.source_account = unhexlify('15d648bfe4d36f196cfb5735ffd8ca54cd4b8233f743f22449de7cf301cdb469')
+        tx.fee = 100
+        tx.sequence_number = 4294967296
```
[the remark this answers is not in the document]
tsusanka: What do you mean by valid? From the Stellar side? Or protobuf maybe? Why 0xffffffff?

tsusanka:
@matejcik I've tried to fix this as per your comments, but to be honest I'm not sure I've done everything as requested -- I got a bit lost in the text.

Could you have a look and edit anything you seem is necessary? I'm not keen on anything, I've basically just wanted to modify the tests, so you can see what is being sent to trezor. If you don't like that we can do it any other way, or actually not do it at all - I'm open to suggestions :).

matejcik:
i think you got it all :) thanks

PR #267, opened 2018-05-23
Add supporting ZenCash

IvanSheigets:


matejcik:
please update the protobuf messages through https://github.com/trezor/trezor-common repo first

matejcik:
also there are no tests and no corresponding PRs to trezor-core/trezor-mcu. This can't be merged on its own. 

I'm going to go ahead and close the PR now; please open a new one when more of the pieces are in place.
As a rule of thumb: new Trezor features can be added to python-trezor only when these features already exist on the Trezor side; usually you want to send PRs simultaneously.

PR #272, opened 2018-06-07
Improve firmware manipulation for Trezor T

### LINKED ISSUE #269: trezorctl: implement firmware_update for both T1 and T2
`firmware_update` function should do the following:

1) use `connect().features.bootloader_mode` to check whether the device is in the bootloader mode first and throw an exception if not

2) use `connect().features.major_version` to determine whether we use T1 or T2

3) use `https://wallet.trezor.io/data/firmware/%d/releases.json` instead of `https://wallet.trezor.io/data/firmware/releases.json` for further processing

matejcik:
* fix #269 by making `firmware_update` more clever
* remove the `--erase` option
* make `wipe_device` bootloader aware (its functionality supersedes `firmware_update --erase`)


prusnak:
utACK 84eb6cb

PR #275, opened 2018-06-11
Stellar: support for get address

tsusanka:
Updates https://github.com/trezor/trezor-core/issues/226 and https://github.com/trezor/trezor-mcu/issues/363


prusnak:
LGTM utACK

trezorlib/client.py
```diff
@@ -1102,6 +1102,10 @@ def self_test(self):
     def stellar_get_public_key(self, address_n):
         return self.call(proto.StellarGetPublicKey(address_n=address_n))
 
+    @expect(proto.StellarAddress)
```
matejcik: please add `@field('address')` above this line, and modify the tests accordingly.

(the equivalent should be done for stellar_get_public_key - if you have the time, please add that too :) )

matejcik:
generally LGTM (apart from the `@field` thing), but there should be a corresponding trezorctl command.
care to add it?
(or i can do it, if you don't mind waiting until (probably) tomorrow)

tsusanka:
I've fixed the `@field` annotation - also for get_public_key. I can add the trezorctl command, but also tomorrow.

prusnak:
Added the trezorctl command as I needed that anyway for testing

matejcik:
perfect, thanks. :+1:

PR #286, opened 2018-07-11
Ripple support

tsusanka:
This introduces support for Ripple, which is coming to trezor-core.

I wasn't sure about the API, so I've done it in the similar way as Stellar's. So the signing command takes a dict as an argument consisting of the fields. I'm not sure it is ideal though, so let me know what you think.


matejcik:
quick notes:
- please don't use `import messages as proto` anymore, i'm getting rid of that weirdness
- please move ripple methods from client to `ripple.py`; take a look at https://github.com/trezor/python-trezor/pull/276, that's how it's supposed to look in the future (except you don't need the `MoveTo` adapters because you're creating new functions)

trezorlib/ripple.py
```diff
@@ -0,0 +1,48 @@
...
+
+import base64
+import struct
+import xdrlib
+
+from . import messages as proto
+
+
+def validate(transaction):
+    if False in (k in transaction for k in ("Fee", "Sequence", "TransactionType", "Amount", "Destination")):
```
matejcik:
```python
if not all(k in transaction for k in (...)):
```
also maybe:
```python
if not all(transaction.get(k) for k in (...)):
```
to check truthiness of the actual values?

trezorlib/ripple.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def create_sign_tx(transaction) -> proto.RippleSignTx:
+    msg = proto.RippleSignTx()
+    msg.fee = transaction["Fee"]
+    msg.sequence = transaction["Sequence"]
+    if "Flags" in transaction:
+        msg.flags = transaction["Flags"]
+    if "LastLedgerSequence" in transaction:
+        msg.last_ledger_sequence = transaction["LastLedgerSequence"]
```
matejcik: instead:
```python
return messages.RippleSignTx(
	fee=transaction["Fee"],
	sequence=transaction["Sequence"],
	flags=transaction.get("Flags"),
	last_ledger_sequence = transaction.get("LastLedgerSequence"),
	payment=create_payment(transaction),
)
```
(perhaps use `get()` for the first two as well, for consistency)

the `dict.get` method will return `None` if key is missing, so you don't need the `if` check.

also maybe `validate(transaction)` within this function?

trezorlib/ripple.py
```diff
@@ -0,0 +1,48 @@
...
+
+    msg.payment = create_payment(transaction)
+    return msg
+
+
+def create_payment(transaction) -> proto.RipplePayment:
+    msg = proto.RipplePayment()
+    msg.amount = transaction["Amount"]
+    msg.destination = transaction["Destination"]
+    return msg
```
matejcik: i'd also put the properties in constructor (same as above), if they are just items from the dict)

trezorlib/client.py
```diff
@@ -1082,6 +1083,21 @@ def self_test(self):
...
+    @expect(proto.RippleAddress)
+    def ripple_get_address(self, address_n, show_display=False):
+        return self.call(
+            proto.RippleGetAddress(
+                address_n=address_n, show_display=show_display))
+
+    @expect(proto.RippleSignedTx)
+    def ripple_sign_tx(self, n, transaction):
+        ripple.validate(transaction)
+        n = self._convert_prime(n)
```
matejcik: no more `_convert_prime` please

matejcik:
re API: where do you expect to get transaction data?
Stellar parses them out of a XDR blob and passes the protobuf objects directly, iirc. That's a reasonable thing to do IMHO. Here, I'd probably prefer that you called `create_sign_tx` outside the signing function, i.e., use `create_sign_tx` to "decode" whatever the actual input data and only then call `ripple.sign_tx`.

I'm probably going to modify Lisk and NEM (which take dicts) to work similarly.

trezorlib/ripple.py
tsusanka: You're right, so your second option it is

trezorlib/ripple.py
tsusanka: Great, fixed

trezorlib/ripple.py
tsusanka: Fixed

tsusanka:
> re API: where do you expect to get transaction data?

Yeah, this is something I wanted to discuss with you. I guess there are three options how the input data can be provided:

1) Because we support only a simple payment scenario, I was considering an option where the user provides the actual arguments directly in a similar way it is for Ethereum. So the command line api would look like this:

`ripple_sign_transaction(self, address_n, destination, amount, fee, sequence, flags=None, last_sequence=None)`

 This has one serious disadvantage that when other transaction types are introduced, we'll have to change this API. So maybe it's not a good idea.

2) Provide the input data as a serialized unsigned transaction. This makes sense, but we'd have to include yet another serialization format - the Ripple's [binary format](https://wiki.ripple.com/Binary_Format), because of course they're not using something common.

3) Provide the input data as a JSON-like dictionary. This will resemble the official Ripple API, see [here](https://developers.ripple.com/payment.html) for an example of a Payment transaction JSON. So my idea is, the user would provide this struct. Because we do not want to parse JSON, it would be a simple dict. That's basically how it is done now.

tsusanka:
> please move ripple methods from client to ripple.py; take a look at #276, that's how it's supposed to look in the future (except you don't need the MoveTo adapters because you're creating new functions)

Ok, I'll move the methods to ripple.py. How do I register it in the client? Should I do something like

```
def ripple_sign_tx(self, n, transaction):
	return ripple.sign_tx(n, transaction)
```

or is it done automatically or ...?

trezorlib/client.py
matejcik: yes, but not here. the `n` should already be the output of `parse_path`, that part is correct

matejcik:
re "register in client" - don't do that at all .) users are expected to `from trezorlib import ripple` and then `ripple.sign_tx(client, ...)`

trezorlib/client.py
tsusanka: Sorry, I'm confused. So what should I use instead of `_convert_prime`?

matejcik:
re API: why not "provide data as protobuf objects"?
then we can have `ripple.binary_to_proto` to decode the binary format, and/or `riple.dict_to_proto` to decode the JSON-like input

tsusanka:
Discussed in person. We'll load the transaction data from a json file and then parse into protobuf. The sign_tx function will accept the protobuf messages

tsusanka:
@matejcik I think this is done from my side. Let me know if I forgot something.

trezorlib/ripple.py
```diff
@@ -0,0 +1,59 @@
...
+
+from . import messages
+from . import tools
+from .client import field
+from .client import expect
+
+
+@field('address')
+@expect(messages.RippleAddress)
+def ripple_get_address(client, address_n, show_display=False):
```
matejcik: please rename to `get_address`

trezorlib/ripple.py
```diff
@@ -0,0 +1,59 @@
...
+    return client.call(
+        messages.RippleGetAddress(
+            address_n=address_n, show_display=show_display))
+
+
+@expect(messages.RippleSignedTx)
+def ripple_sign_tx(client, address_n, msg: messages.RippleSignTx):
```
matejcik: please rename to `sign_tx`

matejcik:
last thing (function names), otherwise LGTM

matejcik:
and done.
thank you, please delete the branch when you're done here :)

PR #287, opened 2018-07-16
Added Zencash supporting

idimon4uk:
Add get_tx test for Zencash
Add test_msg_signtx_zencash

### DIFF
--- a/trezorctl
+++ b/trezorctl
@@ -576,6 +576,10 @@ def sign_tx(connect, coin):
         sequence = click.prompt('Sequence Number to use (RBF opt-in enabled by default)', type=int, default=0xfffffffd)
         script_type = click.prompt('Input type', type=CHOICE_INPUT_SCRIPT_TYPE, default=default_script_type(address_n))
         script_type = script_type if isinstance(script_type, int) else CHOICE_INPUT_SCRIPT_TYPE.typemap[script_type]
+        if txapi.bip115:
+            prev_output = txapi.get_tx(binascii.hexlify(prev_hash).decode("utf-8")).bin_outputs[prev_index]
+            prev_blockhash = prev_output.block_hash
+            prev_blockheight = prev_output.block_height
         inputs.append(proto.TxInputType(
             address_n=address_n,
             prev_hash=prev_hash,
@@ -583,7 +587,14 @@ def sign_tx(connect, coin):
             amount=amount,
             script_type=script_type,
             sequence=sequence,
+            prev_block_hash_bip115=prev_blockhash,
+            prev_block_height_bip115=prev_blockheight,
         ))
+
+    if txapi.bip115:
+        current_block_height = txapi.current_height()
+        block_height = current_block_height - 300  # Zencash recommendation for the better protection
+        block_hash = txapi.get_block_hash(block_height)
 
     outputs = []
     while True:
@@ -604,6 +615,8 @@ def sign_tx(connect, coin):
             address=address,
             amount=amount,
             script_type=script_type,
+            block_hash_bip115=block_hash[::-1],  # Blockhash passed in reverse order
+            block_height_bip115=block_height
         ))
 
     tx_version = click.prompt('Transaction version', type=int, default=2)
--- a/trezorlib/tests/unit_tests/test_tx_api.py
+++ b/trezorlib/tests/unit_tests/test_tx_api.py
@@ -18,9 +18,11 @@ import os
 
 from trezorlib import coins
 from trezorlib import tx_api
+import binascii
 
 TxApiBitcoin = coins.tx_api['Bitcoin']
 TxApiTestnet = tx_api.TxApiInsight("insight_testnet")
+TxApiZencash = coins.tx_api['Zencash']
 
 tests_dir = os.path.dirname(os.path.abspath(__file__))
 
@@ -42,3 +44,13 @@ def test_tx_api_gettx():
 
     TxApiTestnet.get_tx('6f90f3c7cbec2258b0971056ef3fe34128dbde30daa9c0639a898f9977299d54')
     TxApiTestnet.get_tx('d6da21677d7cca5f42fbc7631d062c9ae918a0254f7c6c22de8e8cb7fd5b8236')
+
+
+def test_tx_api_current_block():
+    height = TxApiZencash.current_height()
+    assert height > 347041
+
+
+def test_tx_api_get_block_hash():
+    hash = TxApiZencash.get_block_hash(110000)
+    assert hash == binascii.unhexlify('000000003f5d6ba1385c6cd2d4f836dfc5adf7f98834309ad67e26faef462454')
--- a/trezorlib/tx_api.py
+++ b/trezorlib/tx_api.py
@@ -33,6 +33,18 @@ class TxApi(object):
         url = '%s%s/%s' % (self.url, resource, resourceid)
         return url
 
+    def current_height(self):
+        r = requests.get(self.url + '/status?q=getBlockCount')
+        j = r.json(parse_float=str)
+        block_height = j['info']['blocks']
+        return block_height
+
+    def get_block_hash(self, block_number):
+        r = requests.get(self.url + '/block-index/' + str(block_number))
+        j = r.json(parse_float=str)
+        block_hash = binascii.unhexlify(j['blockHash'])
+        return block_hash
+
     def fetch_json(self, resource, resourceid):
         global cache_dir
         if cache_dir:
@@ -65,11 +77,13 @@ class TxApi(object):
 
 class TxApiInsight(TxApi):
 
-    def __init__(self, network, url=None, zcash=None):
+    def __init__(self, network, url=None, zcash=None, bip115=False):
         super().__init__(network, url)
         self.zcash = zcash
+        self.bip115 = bip115
         if url:
-            self.pushtx_url = url.replace('/api/', '/tx/send')
+            prefix, suffix = url.rsplit('/', maxsplit=1)
+            self.pushtx_url = prefix + '/tx/send'
 
     def get_tx(self, txhash):
 
@@ -97,6 +111,12 @@ class TxApiInsight(TxApi):
             o = t._add_bin_outputs()
             o.amount = int(Decimal(vout['value']) * 100000000)
             o.script_pubkey = binascii.unhexlify(vout['scriptPubKey']['hex'])
+            if self.bip115 and o.script_pubkey[-1] == 0xb4:
+                # Verify if coin implements replay protection bip115 and script includes checkblockatheight opcode. 0xb4 - is op_code (OP_CHECKBLOCKATHEIGHT)
+                # <OP_32> <32-byte block hash> <OP_3> <3-byte block height> <OP_CHECKBLOCKATHEIGHT>
+                tail = o.script_pubkey[-38:]
+                o.block_hash = tail[1:33]  # <32-byte block hash>
+                o.block_height = int.from_bytes(tail[34:37], byteorder='little')  # <3-byte block height>
 
         if self.zcash:
             t.overwintered = data.get('fOverwintered', False)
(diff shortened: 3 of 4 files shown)


trezorlib/tests/device_tests/test_msg_signtx_zencash.py
```diff
@@ -0,0 +1,93 @@
...
+        out1 = proto.TxOutputType(
+            address='znmc8tf4ngHj6YmBr9QNPQMSwu9YwLV6XM6',
+            amount=40000,
+            script_type=proto.OutputScriptType.PAYTOADDRESS,
+            block_hash_bip115=unhexlify('dfdd8aad03f9104a068aae49102926c4d92b1dc38a02b4a471556d2900000000'),
+            block_height_bip115=291331,
+        )
+
+        with self.client:
+            self.client.set_expected_responses([
```
matejcik: Is there any reason to check responses one by one? Does zencash change signature flow from the default btc code -- IOW, would the list of responses be different if you were testing a different coin?

if not, please drop the whole `with` block and just run the `sign_tx` function

trezorlib/tests/unit_tests/test_tx_api.py
```diff
@@ -21,6 +21,7 @@
 
 TxApiBitcoin = coins.tx_api['Bitcoin']
 TxApiTestnet = tx_api.TxApiInsight("insight_testnet")
+TxApiZencash = coins.tx_api['Zencash']
```
matejcik: is there a reason to test the tx api here? if you want to have comprehensive test, you should add tests for your `get_block` method (but see other comments)

trezorlib/tx_api.py
```diff
@@ -65,11 +76,15 @@ def get_tx(self, txhash):
...
+    def __init__(self, network, url=None, zcash=None, bip115=False):
         super().__init__(network, url)
         self.zcash = zcash
+        self.bip115 = bip115
         if url:
-            self.pushtx_url = url.replace('/api/', '/tx/send')
+            http_protocol, link = url.split('://')
+            if ('/' in link):
+                link, api = link.split('/')
+            self.pushtx_url = http_protocol + '://' + link + '/tx/send'
```
matejcik: no.

What are you trying to accomplish here exactly? Is it that your api server doesn't contain `/api/` in the URL?
I can see that this code is currently broken anyway, because there are no trailing slashes.

Why don't you first strip out trailing `"/api"` if present, and then append `"/tx/send"` always? This splitting on protocol makes no sense.

trezorlib/tx_api.py
```diff
@@ -97,6 +112,13 @@ def get_tx(self, txhash):
             o = t._add_bin_outputs()
             o.amount = int(Decimal(vout['value']) * 100000000)
             o.script_pubkey = binascii.unhexlify(vout['scriptPubKey']['hex'])
+            if(self.bip115 and len(o.script_pubkey) == 63 and o.script_pubkey[-1:] == b'\xb4'):
+                o.block_hash = o.script_pubkey[-37:-5]
+                o.block_height = int.from_bytes(o.script_pubkey[-4:-1], byteorder='little')
+
+            else:
+                o.block_hash = None
+                o.block_height = None
```
matejcik: this `else` here is a no-op (all fields are preinitialized to `None` anyway) so please remove it

trezorlib/tx_api.py
```diff
@@ -97,6 +112,13 @@ def get_tx(self, txhash):
             o = t._add_bin_outputs()
             o.amount = int(Decimal(vout['value']) * 100000000)
             o.script_pubkey = binascii.unhexlify(vout['scriptPubKey']['hex'])
+            if(self.bip115 and len(o.script_pubkey) == 63 and o.script_pubkey[-1:] == b'\xb4'):
+                o.block_hash = o.script_pubkey[-37:-5]
+                o.block_height = int.from_bytes(o.script_pubkey[-4:-1], byteorder='little')
```
matejcik: again, what is this magic exactly?

1. don't wrap the if-condition in parentheses
2. to check for the last byte of scriptPubKey, do this: `.script_pubkey[-1] == 0xb4`
3. please add a comment explaining the numbers. perhaps a link to some spec would be helpful. ideally also describe the structure you're accessing.
4. is there a reason to count backwards? you are enforcing a fixed length of the script anyway.

trezorctl
```diff
@@ -599,11 +604,17 @@ def sign_tx(connect, coin):
         amount = click.prompt('Amount to spend (satoshis)', type=int)
         script_type = click.prompt('Output type', type=CHOICE_OUTPUT_SCRIPT_TYPE, default=default_script_type(address_n))
         script_type = script_type if isinstance(script_type, int) else CHOICE_OUTPUT_SCRIPT_TYPE.typemap[script_type]
+        if txapi.bip115:
+            (block_hash, block_height) = txapi.get_block(300)
```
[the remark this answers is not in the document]
matejcik: also: this whole thing looks wrong. why **300** in particular? Will this be the same in all BIP-115 coins? I don't think so.

Also you're using a brand new `get_block` function which is questionable in the first place. Can't you reuse block heights from the inputs?

trezorlib/tx_api.py
```diff
@@ -33,6 +33,17 @@ def get_url(self, resource, resourceid):
         url = '%s%s/%s' % (self.url, resource, resourceid)
         return url
 
+    def get_block(self, block_number):  # currunt block - block_number
```
matejcik: this whole function is highly questionable and I would prefer if it didn't exist, and you extracted the target block height from somewhere else

matejcik:
hi, i have reviewed your changes. Please update the code and provide explanations where appropriate.

trezorctl
```diff
@@ -576,13 +576,18 @@ def sign_tx(connect, coin):
         sequence = click.prompt('Sequence Number to use (RBF opt-in enabled by default)', type=int, default=0xfffffffd)
         script_type = click.prompt('Input type', type=CHOICE_INPUT_SCRIPT_TYPE, default=default_script_type(address_n))
         script_type = script_type if isinstance(script_type, int) else CHOICE_INPUT_SCRIPT_TYPE.typemap[script_type]
+        prev_output = txapi.get_tx(str(binascii.hexlify(prev_hash))[2:-1]).bin_outputs[prev_index]
```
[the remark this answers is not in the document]
idimon4uk: Sorry, It was my fault, fixed.

trezorlib/tests/device_tests/test_msg_signtx_zencash.py
idimon4uk: Yes , you're right, it's no reason to do this validation, I will delete it.

trezorlib/tx_api.py
matejcik: just noticed now: this extraction of `block_hash` and `block_height` only works with `o.script_pubkey`.

That means that it doesn't need to be done here. Instead it should be done in Trezor itself, and the new fields should be removed from `InputType`

trezorctl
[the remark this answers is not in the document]
matejcik: also, the values are the same for all outputs.

so please move the call out of the `while True` output loop and reuse the values

trezorlib/tx_api.py
```diff
@@ -33,6 +33,17 @@ def get_url(self, resource, resourceid):
...
+        r = requests.get(self.url + '/blocks/')
+        j = r.json(parse_float=str)
+        block_height = j['blocks'][0]['height']
+        block_height -= block_number
+        r = requests.get(self.url + '/block-index/' + str(block_height))
+        j = r.json(parse_float=str)
+        block_hash = j['blockHash']
+        block_hash_flipped = binascii.unhexlify("".join(reversed([block_hash[i:i + 2] for i in range(0, len(block_hash), 2)])))
```
matejcik: instead:
```python
block_hash = binascii.unhexlify(j["blockHash"])
return block_hash[::-1], block_height
```

trezorlib/tx_api.py
matejcik: as discussed on gitter: instead rename the function to, maybe, `bip115_get_nth_block`
and add a comment or docstring that explains that this gets the Nth block from top and returns its reversed hash and block height

prusnak: I would expect that `get_block(X)` call returns block with block_height==X, let's do it this way. Also the function does not return block, just it hash, so it should be called `get_blockhash(X)`.

To implement the intended behavior, let's create `get_blockheight()` function which returns the current blockheight and the desired outcome would be:

```
current = api.get_blockheight()
h = get_blockhash(current - 300)  # say in comment here why 300
```

Also not sure about `get_` prefix, maybe the functions should be called just `blockhash` and `blockheight`, but I leave that to @matejcik.

matejcik: yes, this is a good proposal. also the flipping should be done in client code, i.e., in `trezorctl`

as for naming, I don't particularly care, but my choice would be:
`current_height()` for the height function (it looks like a "property" of the API, which seems appropriate)
`get_block_hash(i)` for the other one

idimon4uk:
Hello, Please review the code modifications made. We appreciate your feedback.

trezorlib/tests/unit_tests/test_tx_api.py
```diff
@@ -42,3 +43,11 @@ def test_tx_api_gettx():
...
     TxApiTestnet.get_tx('6f90f3c7cbec2258b0971056ef3fe34128dbde30daa9c0639a898f9977299d54')
     TxApiTestnet.get_tx('d6da21677d7cca5f42fbc7631d062c9ae918a0254f7c6c22de8e8cb7fd5b8236')
+
+
+def test_tx_api_current_block():
+    TxApiZencash.current_height()
+
+
+def test_tx_api_get_block_hash():
+    TxApiZencash.get_block_hash(110000)  # 110000 - test value
```
matejcik: your tests need some asserts

...i realize the previous test doesn't have any, and it's a bad test for that

you should do something like:
```python
def test_tx_api_current_block():
￼    height = TxApiZencash.current_height()
	assert height > 12345689  # replace this number with today's block height.
￼ 

￼def test_tx_api_get_block_hash():
￼    hash = TxApiZencash.get_block_hash(110000)
	assert hash == binascii.hexlify("klsjadféjěšklasdfasjlkjsa") # replace with block hash 110000
```

you should also add a test for the bip115 fields: make a function `test_tx_api_bip115`, use `tx = TxApiZencash.get_tx("lakdfoiaslwltkj")`, and then `assert` that the fields (`block_height`, `block_hash`) contain correct values.

trezorlib/tx_api.py
```diff
@@ -97,6 +111,12 @@ def get_tx(self, txhash):
             o = t._add_bin_outputs()
             o.amount = int(Decimal(vout['value']) * 100000000)
             o.script_pubkey = binascii.unhexlify(vout['scriptPubKey']['hex'])
+            if self.bip115 and o.script_pubkey[-1] == 0xb4:  # Verify if coin implements replay protection bip115 and script includes checkblockatheight opcode. 0xb4 - is op_code (OP_CHECKBLOCKATHEIGHT)
+                # For exapmle: 76a914b93896509fde251ea0f93cc9f32fa9eb010956b988ac201341e03aaa53e209874363bf85b5cbedf21f12443833f9f721eb500a0000000003943c05b4 bytes we have recieved
```
matejcik: please keep the comment after `if` (and move it to next line), but remove the examples.
instead, please describe the structure, like:
"<original scriptPubKey> <OP_32> <32-byte block hash> <OP_3> <3-byte block height> <OP_CHECKBLOCKATHEIGHT>"
(I'm not sure if I wrote it correctly - If I made a mistake, please write the correct structure :) )

also, I see you removed the length check. In that case I think it's better to extract the "tail" first, and then count forward
```python
tail = o.script_pubkey[-38:]
o.block_hash = tail[1:33]
o.block_height = tail[34:37]
```

matejcik:
Thanks for your work, this is looking much better.
I still have some comments, please fix the remaining problems.

idimon4uk:
@matejcik Please take a look on the last changes

matejcik:
this is looking good.
thank you!

PR #297, opened 2018-08-14
add more eip155 test cases

hackmod:
* add more eip155 test cases for `chain_id` >= `MAX_CHAIN_ID(2147483630)`
* Pirl `chain_id` case (`3125659152`)
* `chain_id` == `MAX_UINT32(4294967295)` case.

see also https://github.com/trezor/trezor-mcu/pull/399


hackmod:
tested on window 10
image

PR #300, opened 2018-08-18
Add cardano support to trezorctl and some tests

ddeath:
Added device tests for get_address and sign_transaction calls.

To the trezorctl was also added get_public_key, sign and verify message

### DIFF
--- a/trezorctl
+++ b/trezorctl
@@ -32,6 +32,7 @@ import click
 
 from trezorlib import (
     btc,
+    cardano,
     coins,
     cosi,
     debuglink,
@@ -1077,6 +1078,72 @@ def ethereum_sign_tx(
     else:
         return "Signed raw transaction: %s" % tx_hex
 
+
+#
+# ADA functions
+#
+
+@cli.command(help='Sign Cardano transaction.')
+@click.option('-f', '--file', type=click.File('r'), required=True, help='Transaction in JSON format')
+@click.pass_obj
+def cardano_sign_tx(connect, file):
+    client = connect()
+
+    transaction = json.load(file)
+
+    inputs = [cardano.create_input(input) for input in transaction['inputs']]
+    outputs = [cardano.create_output(output) for output in transaction['outputs']]
+    transactions = transaction['transactions']
+
+    signed_transaction = cardano.sign_tx(client, inputs, outputs, transactions)
+
+    return {
+        'tx_hash': binascii.hexlify(signed_transaction.tx_hash).decode(),
+        'tx_body': binascii.hexlify(signed_transaction.tx_body).decode()
+    }
+
+
+@cli.command(help='Get Cardano address.')
+@click.option('-n', '--address', required=True, help="BIP-32 path to key, e.g. m/44'/1815'/0'/0/0")
+@click.option('-d', '--show-display', is_flag=True)
+@click.pass_obj
+def cardano_get_address(connect, address, show_display):
+    client = connect()
+    address_n = tools.parse_path(address)
+
+    return cardano.get_address(client, address_n, show_display)
+
+
+@cli.command(help='Get Cardano public key.')
+@click.option('-n', '--address', required=True, help="BIP-32 path to key, e.g. m/44'/1815'/0'/0/0")
+@click.pass_obj
+def cardano_get_public_key(connect, address):
+    client = connect()
+    address_n = tools.parse_path(address)
+
+    return cardano.get_public_key(client, address_n)
+
+
+@cli.command(help='Sign Cardano message.')
+@click.option('-n', '--address', required=True, help="BIP-32 path to key, e.g. m/44'/1815'/0'/0/0")
+@click.option('-m', '--message', required=True, help="String message to sign")
+@click.pass_obj
+def cardano_sign_message(connect, address, message):
+    client = connect()
+    address_n = tools.parse_path(address)
+
+    return cardano.sign_message(client, address_n, message)
+
+
+@cli.command(help='Verify Cardano message.')
+@click.option('-p', '--public_key', required=True, help="Public key as hex string")
+@click.option('-s', '--signature', required=True, help="Signature as hex string")
+@click.option('-m', '--message', required=True, help="String message which was signed")
+@click.pass_obj
+def cardano_verify_message(connect, public_key, signature, message):
+    client = connect()
+
+    return cardano.verify_message(client, public_key, signature, message)
 
 #
 # NEM functions
--- a/trezorctl
+++ b/trezorctl
@@ -1083,29 +1083,38 @@ def ethereum_sign_tx(
 # ADA functions
 #
 
-@cli.command(help='Sign Cardano transaction.')
-@click.option('-f', '--file', type=click.File('r'), required=True, help='Transaction in JSON format')
+
+@cli.command(help="Sign Cardano transaction.")
+@click.option(
+    "-f",
+    "--file",
+    type=click.File("r"),
+    required=True,
+    help="Transaction in JSON format",
+)
 @click.pass_obj
 def cardano_sign_tx(connect, file):
     client = connect()
 
     transaction = json.load(file)
 
-    inputs = [cardano.create_input(input) for input in transaction['inputs']]
-    outputs = [cardano.create_output(output) for output in transaction['outputs']]
-    transactions = transaction['transactions']
+    inputs = [cardano.create_input(input) for input in transaction["inputs"]]
+    outputs = [cardano.create_output(output) for output in transaction["outputs"]]
+    transactions = transaction["transactions"]
 
     signed_transaction = cardano.sign_tx(client, inputs, outputs, transactions)
 
     return {
-        'tx_hash': binascii.hexlify(signed_transaction.tx_hash).decode(),
-        'tx_body': binascii.hexlify(signed_transaction.tx_body).decode()
+        "tx_hash": binascii.hexlify(signed_transaction.tx_hash).decode(),
+        "tx_body": binascii.hexlify(signed_transaction.tx_body).decode(),
     }
 
 
-@cli.command(help='Get Cardano address.')
-@click.option('-n', '--address', required=True, help="BIP-32 path to key, e.g. m/44'/1815'/0'/0/0")
-@click.option('-d', '--show-display', is_flag=True)
+@cli.command(help="Get Cardano address.")
+@click.option(
+    "-n", "--address", required=True, help="BIP-32 path to key, e.g. m/44'/1815'/0'/0/0"
+)
+@click.option("-d", "--show-display", is_flag=True)
 @click.pass_obj
 def cardano_get_address(connect, address, show_display):
     client = connect()
@@ -1114,8 +1123,10 @@ def cardano_get_address(connect, address, show_display):
     return cardano.get_address(client, address_n, show_display)
 
 
-@cli.command(help='Get Cardano public key.')
-@click.option('-n', '--address', required=True, help="BIP-32 path to key, e.g. m/44'/1815'/0'/0/0")
+@cli.command(help="Get Cardano public key.")
+@click.option(
+    "-n", "--address", required=True, help="BIP-32 path to key, e.g. m/44'/1815'/0'/0/0"
+)
 @click.pass_obj
 def cardano_get_public_key(connect, address):
     client = connect()
@@ -1124,9 +1135,11 @@ def cardano_get_public_key(connect, address):
     return cardano.get_public_key(client, address_n)
 
 
-@cli.command(help='Sign Cardano message.')
-@click.option('-n', '--address', required=True, help="BIP-32 path to key, e.g. m/44'/1815'/0'/0/0")
-@click.option('-m', '--message', required=True, help="String message to sign")
+@cli.command(help="Sign Cardano message.")
+@click.option(
+    "-n", "--address", required=True, help="BIP-32 path to key, e.g. m/44'/1815'/0'/0/0"
+)
+@click.option("-m", "--message", required=True, help="String message to sign")
 @click.pass_obj
 def cardano_sign_message(connect, address, message):
     client = connect()
@@ -1135,15 +1148,16 @@ def cardano_sign_message(connect, address, message):
     return cardano.sign_message(client, address_n, message)
 
 
-@cli.command(help='Verify Cardano message.')
-@click.option('-p', '--public_key', required=True, help="Public key as hex string")
-@click.option('-s', '--signature', required=True, help="Signature as hex string")
-@click.option('-m', '--message', required=True, help="String message which was signed")
+@cli.command(help="Verify Cardano message.")
+@click.option("-p", "--public_key", required=True, help="Public key as hex string")
+@click.option("-s", "--signature", required=True, help="Signature as hex string")
+@click.option("-m", "--message", required=True, help="String message which was signed")
 @click.pass_obj
 def cardano_verify_message(connect, public_key, signature, message):
     client = connect()
 
     return cardano.verify_message(client, public_key, signature, message)
+
 
 #
 # NEM functions
--- a/trezorlib/cardano.py
+++ b/trezorlib/cardano.py
@@ -30,6 +30,7 @@ from .client import session
 REQUIRED_FIELDS_TRANSACTION = ('inputs', 'outputs', 'transactions')
 REQUIRED_FIELDS_INPUT = ('path', 'prev_hash', 'prev_index', 'type')
 
+
 @field('address')
 @expect(messages.CardanoAddress)
 def get_address(client, address_n, show_display=False):
--- a/trezorlib/cardano.py
+++ b/trezorlib/cardano.py
@@ -14,45 +14,32 @@
 # You should have received a copy of the License along with this library.
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
-import base64
-import struct
 import binascii
 from typing import List
 
-from . import messages
-from . import tools
-from .client import CallException
-from .client import field
-from .client import expect
-from .client import session
+from . import messages, tools
+from .tools import CallException, expect, session
+
+REQUIRED_FIELDS_TRANSACTION = ("inputs", "outputs", "transactions")
+REQUIRED_FIELDS_INPUT = ("path", "prev_hash", "prev_index", "type")
 
 
-REQUIRED_FIELDS_TRANSACTION = ('inputs', 'outputs', 'transactions')
-REQUIRED_FIELDS_INPUT = ('path', 'prev_hash', 'prev_index', 'type')
-
-
-@field('address')
-@expect(messages.CardanoAddress)
+@expect(messages.CardanoAddress, field="address")
 def get_address(client, address_n, show_display=False):
     return client.call(
-        messages.CardanoGetAddress(
-            address_n=address_n, show_display=show_display))
+        messages.CardanoGetAddress(address_n=address_n, show_display=show_display)
+    )
 
 
 @expect(messages.CardanoPublicKey)
 def get_public_key(client, address_n):
-    return client.call(
-        messages.CardanoGetPublicKey(address_n=address_n)
-    )
+    return client.call(messages.CardanoGetPublicKey(address_n=address_n))
 
 
 @expect(messages.CardanoMessageSignature)
 def sign_message(client, address_n, message):
     return client.call(
-        messages.CardanoSignMessage(
-            address_n=address_n,
-            message=message.encode()
-        )
+        messages.CardanoSignMessage(address_n=address_n, message=message.encode())
     )
 
 
@@ -62,7 +49,7 @@ def verify_message(client, public_key, signature, message):
             messages.CardanoVerifyMessage(
                 public_key=binascii.unhexlify(public_key),
                 signature=binascii.unhexlify(signature),
-                message=message.encode()
+                message=message.encode(),
             )
         )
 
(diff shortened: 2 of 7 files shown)


tsusanka:
@matejcik I can leave this for you to review, right?

trezorlib/cardano.py
```diff
@@ -0,0 +1,130 @@
...
+        messages.CardanoVerifyMessage(
+            public_key=binascii.unhexlify(public_key),
+            signature=binascii.unhexlify(signature),
+            message=message.encode()
+        )
+    )
+
+
+@session
+def cardano_sign_transaction(client, transaction):
```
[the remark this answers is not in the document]
matejcik: also the signature is not great, see comments in `trezorctl` and on `create_sign_transaction_message`.

In general, these functions should consume either simple values, or filled out protobuf messages. In this case, best I could come up with is:
```
def sign_tx(client, inputs: List[CardanoTxInputType], outputs: List[CardanoTxOutputType], transactions: List[bytes])
```

trezorlib/cardano.py
```diff
@@ -0,0 +1,130 @@
...
+    msg = create_sign_transaction_message(transaction)
+
+    response = client.call(msg)
+    while response.tx_index is not None:
+        tx_index = response.tx_index
+
+        ack_message = create_tx_ack_message(transaction.get('transactions'), tx_index)
```
matejcik: the `create_tx_ack_message` is sort of pointless, esp. if you get `transactions` directly as an argument. please inline it here.

(also maybe verify that the requested index makes sense? but AFAICT the `index` is pointless anyway, so maybe it doesn't matter)

trezorlib/tests/device_tests/test_msg_cardano_get_address.py
```diff
@@ -0,0 +1,52 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the License along with this library.
+# If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
+
+import pytest
+
+from .common import TrezorTest
+from .conftest import TREZOR_VERSION
```
matejcik: you don't seem to be using `TREZOR_VERSION`

trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
```diff
@@ -0,0 +1,137 @@
...
+from trezorlib import messages
+from trezorlib.cardano import cardano_sign_transaction, create_sign_transaction_message, create_tx_ack_message
+
+
+@pytest.mark.cardano
+@pytest.mark.skip_t1  # T1 support is not planned
+class TestMsgCardanoSignTransaction(TrezorTest):
+
+    def test_cardano_sign_transaction(self):
+        self.setup_mnemonic_nopin_nopassphrase()
```
matejcik: ISTM this test could be vastly simplified if you used `set_expected_responses`

see e.g. https://github.com/trezor/python-trezor/blob/master/trezorlib/tests/device_tests/test_msg_signtx.py

matejcik:
I have reviewed the changes. Please see comments and make modifications where appropriate.

In general, my main question is, how do you expect this to be used, from the user side? Is there a tool/service that generates the appropriate JSON data? If not, then, I suppose we'll keep it like this for now. I'm working on some changes that would make it more useful; namely, converting a dict to protobuf directly.

In any case, unless there are good reasons against, please go with my suggestion for the `sign_tx()` signature. Then it depends on where the JSON comes from. If the format is arbitrary, I suggest parsing it in `trezorctl`:
```
inputs = [cardano.create_input(input) for input in transaction['inputs']]
outputs = [cardano.create_output(output) for outputs in transaction['outputs']]
transactions = transaction['transactions']
```
If the format comes from somewhere else, I'd keep the `create_transaction` function, but make it return a tuple `inputs, outputs, transactions` as opposed to a protobuf message (which doesn't include the "transactions" field)`

matejcik:
Also please add the remaining tests (sign/verify message, get public key i think?)

trezorlib/cardano.py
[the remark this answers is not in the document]
ddeath: fixed

trezorctl
```diff
@@ -890,6 +891,62 @@ def ethereum_sign_tx(connect, host, chain_id, address, value, gas_limit, gas_pri
...
+#
+# ADA functions
+#
+
+@cli.command(help='Sign Cardano transaction.')
+@click.option('-f', '--file', type=click.File('r'), required=True, help='Transaction in JSON format')
+@click.pass_obj
+def cardano_sign_transaction(connect, file):
+    client = connect()
+    signed_transaction = cardano.cardano_sign_transaction(client, transaction=json.load(file))
```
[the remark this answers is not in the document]
ddeath: fixed with you metioned suggestion bellow

trezorlib/cardano.py
```diff
@@ -0,0 +1,130 @@
...
+            public_key=binascii.unhexlify(public_key),
+            signature=binascii.unhexlify(signature),
+            message=message.encode()
+        )
+    )
+
+
+@session
+def cardano_sign_transaction(client, transaction):
+    msg = create_sign_transaction_message(transaction)
```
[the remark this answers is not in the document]
ddeath: fixed

trezorlib/cardano.py
```diff
@@ -0,0 +1,130 @@
...
+
+    response = client.call(msg)
+    while response.tx_index is not None:
+        tx_index = response.tx_index
+
+        ack_message = create_tx_ack_message(transaction.get('transactions'), tx_index)
+        response = client.call(ack_message)
+
+    return {
```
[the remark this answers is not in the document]
ddeath: It is so user will see what is hash and what is body without looking at code. if it is only tuple you will have hard times to identify body vs hash as both looks same

trezorlib/cardano.py
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+
+def create_tx_ack_message(transactions, index):
+    transaction_data = binascii.unhexlify(transactions[index])
+    return messages.CardanoTxAck(transaction=transaction_data)
+
+
+def create_sign_transaction_message(transaction) -> messages.CardanoSignTransaction:
+    if not all(transaction.get(k) for k in ('inputs', 'outputs', 'transactions')):
```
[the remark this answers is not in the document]
ddeath: Fixed except output fields as there is only amount required and one other but it does not matter which one - well there must be at least one address_n (at least in one of the outputs)

trezorlib/cardano.py
```diff
@@ -0,0 +1,130 @@
...
+                address_n=tools.parse_path(path),
+                prev_hash=binascii.unhexlify(input.get('prev_hash')),
+                prev_index=input.get('prev_index'),
+                type=input.get('type')
+            )
+        )
+
+    outputs = []
+    for output in transaction.get('outputs'):
+        if not output.get('amount') or not(output.get('address') or output.get('path')):
```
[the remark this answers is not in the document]
ddeath: Not sure how could I use `all` whene there are `not || not ( || )` which is same as `not || (not && not)`

trezorlib/cardano.py
```diff
@@ -0,0 +1,130 @@
...
+def sign_message(client, address_n, message):
+    return client.call(
+        messages.CardanoSignMessage(
+            address_n=address_n,
+            message=message.encode()
+        )
+    )
+
+
+def verify_message(client, public_key, signature, message):
```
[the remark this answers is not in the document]
ddeath: fixed

trezorlib/cardano.py
ddeath: fixed without index verification

trezorlib/tests/device_tests/test_msg_cardano_get_address.py
ddeath: removed

trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
ddeath: I want to used that but I was unsure how to use it when I need to do swipes

trezorlib/cardano.py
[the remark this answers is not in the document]
matejcik: this is not user-facing function though. please return a tuple and do the conversion in `trezorctl`

trezorlib/cardano.py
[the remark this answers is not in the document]
matejcik: ok, my bad, didn't notice the nesting

trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
matejcik: Ah, the swipes thing again! Yes, it makes sense in that case. We need a better infrastructure for simulating swipes.

ddeath:
@matejcik I tried to fix all your comments and also added tests...

About that JSON tool - there is no such a tool. If somebody want to use it then they could look here:
https://github.com/vacuumlabs/cardanolite/blob/78b03bbcf80a9b4d04444d0f954c7c8184ea636d/app/frontend/wallet/cardano-wallet.js#L82

There is our own code for creating tx or they can take a look to connect v5.

The CI is failing because of some other test fails to fetch some data from internet - I guess that we did not break it

trezorlib/cardano.py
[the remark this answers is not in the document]
ddeath: fixed

trezorlib/cardano.py
```diff
@@ -0,0 +1,127 @@
...
+    ))
+
+    while response.tx_index is not None:
+        tx_index = response.tx_index
+
+        transaction_data = binascii.unhexlify(transactions[tx_index])
+        ack_message = messages.CardanoTxAck(transaction=transaction_data)
+        response = client.call(ack_message)
+
+    return (
```
matejcik:
```
return response.tx_hash, response.tx_body
```

trezorctl
```diff
@@ -890,6 +891,72 @@ def ethereum_sign_tx(connect, host, chain_id, address, value, gas_limit, gas_pri
...
+    inputs = [cardano.create_input(input) for input in transaction['inputs']]
+    outputs = [cardano.create_output(output) for output in transaction['outputs']]
+    transactions = transaction['transactions']
+
+    signed_transaction = cardano.sign_tx(client, inputs, outputs, transactions)
+
+    return {
+        'tx_hash': signed_transaction[0],
+        'tx_body': signed_transaction[1]
+    }
```
matejcik:
```
tx_hash, tx_body = cardano.sign_tx(...)

return {
	"tx_hash": binascii.hexlify(tx_hash).decode(),
	"tx_body": binascii.hexlify(tx_body).decode(),
}
```

matejcik:
Tests look reasonable. Please fix my latest comments, I'll look over the whole thing again tomorrow.

Travis failure is on our side; if you rebase on top of current master, it should go away.

trezorlib/cardano.py
ddeath: I just returned response because it is CardanoSignedTx message now. Is that ok?

ddeath:
I fixed comments and also did the rebase.

tsusanka:
@ddeath we would like to unify what derivation paths are sent to which coins. We came up with this [doc](https://github.com/trezor/trezor-core/blob/master/docs/coins/README.md). We would like Cardano wallets to send `44'/1815'/a'` for get public key and `44'/1815'/a'/0/i` for get_address / sign.

1) Is that okay on your side?
2) Do you think you could modify the tests accordingly? I know it's a pain to regenerate the data, but the tests serve as a point of reference sometimes and it would be unfortunate to have it incorrect.

ddeath:
@tsusanka 

1. I would suggest to change get_address and sign to `44'/1815'/a'/{0,1}/i` as I have almost ready PR for adding support for cardano derivation scheme v2 which uses 0 or 1 in that place. In the v1 it must be 0 but in v2 it can be 0 or 1
2. I will regenerate the tests

matejcik:
I finally got to review the current state. It looks mostly OK.

Two things:

1. please add at least one invalid sig to verify_message test :) this way it's not testing much.

2. do the tests pass in the current `trezor-core` master? if not, they should all be marked `xfail`

tsusanka:
> I would suggest to change get_address and sign to 44'/1815'/a'/{0,1}/i as I have almost ready PR for adding support for cardano derivation scheme v2 which uses 0 or 1 in that place. In the v1 it must be 0 but in v2 it can be 0 or 1

Great, agreed.

ddeath:
@tsusanka I changed test so it reflect the derivation paths. Please check if that is what you mean. Also I overlooked the get_address and sign_transaction derivation paths and would suggest to change it to `44'/1815'/a'/{0,1}/{i, i'}` as the official cardano wallet Daedalus is working with addresses derived with `i'` it is also possible to use `i` but for example Binance has problems to accept these kind of addresses.
Also in derivation scheme v2 is used `i` by default in cardano new wallet Icarus.

@matejcik added invalid signatures tests. The test should be passing on master, but they are not because there was introduced bugs couple days ago in this [commit](https://github.com/trezor/trezor-core/commit/8421fa4fdb0f9b08914a326953be34d500c86460#diff-2031ecc0d88f4ccde1228075b9452161R9) where the name of messages are incorrectly swaped and also in this [commit](https://github.com/trezor/trezor-core/commit/c1b284cdfa22126b566c0eaafa9257feeb23f261#diff-ababffd1a91b14b4994ccc5d89b283cfR10) where the `.ui` was not renamed to `.layout`

So not sure if I should mark it as xfail. I will create a PR for adding derivation scheme v2 in couple of hours where I fixed it already. But not sure about your workflow so please just point me out what is the way to do it in this situation.

Thanks

matejcik:
Btw for further reference it would be great if you didn't squash the commit every time, so we could see changes from last time and not have to read through the whole thing.

If your code is in core, and the tests are *supposed* to pass, then no xfail is needed. xfail is for cases when the tests are submitted ahead of the device code.

trezorlib/tests/device_tests/test_msg_cardano_sign_message.py
```diff
@@ -0,0 +1,57 @@
...
+
+from .common import TrezorTest
+from binascii import hexlify
+from trezorlib.cardano import sign_message
+from trezorlib.tools import parse_path
+
+
+@pytest.mark.cardano
+@pytest.mark.skip_t1  # T1 support is not planned
+class TestMsgCardanoGetPublicKey(TrezorTest):
```
matejcik: please rename to match test contents

trezorlib/tests/device_tests/test_msg_cardano_verify_message.py
```diff
@@ -0,0 +1,49 @@
...
+
+import pytest
+
+from .common import TrezorTest
+from trezorlib.cardano import verify_message
+
+
+@pytest.mark.cardano
+@pytest.mark.skip_t1  # T1 support is not planned
+class TestMsgCardanoGetPublicKey(TrezorTest):
```
matejcik: please rename test and method names, and consider using `pytest.mark.parametrize`

trezorlib/tests/device_tests/test_msg_cardano_sign_message.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_cardano_get_public_key(self):
```
matejcik: also this

while you're at it, consider using `@pytest.mark.parametrize` to simplify test code here, see e.g. `test_cancel`

trezorlib/tests/device_tests/test_msg_cardano_sign_message.py
```diff
@@ -0,0 +1,57 @@
...
+        self.client.load_device_by_mnemonic(
+            mnemonic='plastic that delay conduct police ticket swim gospel intact harsh obtain entire',
```
matejcik: Any reason you're using this mnemonic? If not, please regenerate the sigs with the default `all all all...`

(a good reason would be, e.g., if these were officially documented test vectors)

trezorlib/tests/device_tests/test_msg_cardano_sign_message.py
```diff
@@ -0,0 +1,57 @@
...
+            ('New Test message to sign', "m/44'/1815'/0'/0/1", '10fa0ba6705a34e8a6f4af77fb20939a5102e7391fcd26e038ae387e91152b49c26d4d6a1f270a84bd04473f4689e06368997ba740dc2b35fddfe7a07de1a307'),
+            ('Another Test message to sign', "m/44'/1815'/0'/0/2", '72d0b49f072b9c4cf3d034e2ef602b4fe5a601423f3340aa91c23809675442354bfe9a8bf0f127416e79229931b4261709e1f2d513cf596835ce83939aa24108'),
+            ('Just another Test message to sign', "m/44'/1815'/0'/0/3", '796b66b54f79816bffa5ae48eb7bbbddce96c4a40a85e5aa9fb17fe30b0da8b689a62d47897b6471b6d595a5180c9cbd9c8c6c78486e214e0d759b17681ff404'),
+        ]
+
+        for (message, derivation_path, expected_signature) in correct_messages:
+            signature = sign_message(self.client, parse_path(derivation_path), message)
+            assert expected_signature == hexlify(signature.signature).decode('utf8')
+
+        incorrect_messages = [
```
matejcik: FWIW you don't need to test invalid sigs here, because you're already testing against an expected value; it's very unlikely that a `sign_message` would produce the particular signature that you're testing it's _not producing_.

(as opposed to the `verify_message` method, where you need a test that it _can_ fail, that it doesn't return `True` for every input)

unless of course these are vectors for some known failure mode

trezorlib/cardano.py
```diff
@@ -0,0 +1,124 @@
...
+from .client import field
+from .client import expect
+from .client import session
+
+
+REQUIRED_FIELDS_TRANSACTION = ('inputs', 'outputs', 'transactions')
+REQUIRED_FIELDS_INPUT = ('path', 'prev_hash', 'prev_index', 'type')
+
+
+@expect(messages.CardanoAddress)
```
matejcik: this should probably extract the `@field("address")` directly, right?

trezorlib/tests/device_tests/test_msg_cardano_sign_message.py
matejcik: In that case let's change it to the default `all all all...`. It should be changed in trezor-core too, but we'll probably do that later along with other currencies that use custom mnemonics for tests.

trezorlib/tests/device_tests/test_msg_cardano_get_address.py
```diff
@@ -25,14 +25,14 @@
...
         # data from https://iancoleman.io/bip39/#english
         self.setup_mnemonic_nopin_nopassphrase()
 
-        address = get_address(self.client, parse_path("m/44'/1815'/0'/0/0")).address
-        assert address == '2w1sdSJu3GVfqnGAyqAdaWrN8Txv1vCZTN1Pe2AA54ysjWNbNzma3WVtSJfMc6HpM9KEQsdJ7oALPwfQWesRp8QDsFRQpzuNrdq'
-        address = get_address(self.client, parse_path("m/44'/1815'/0'/0/1")).address
-        assert address == '2w1sdSJu3GVhMmEYeGYEPWahV1V17pFw59GfgqjSRqa6x1rKFxbyCZrQWLe78xdSx3zyed6DrrN5yMgoY7ST2vJeaMzUDB7W3WG'
-        address = get_address(self.client, parse_path("m/44'/1815'/0'/0/2")).address
-        assert address == '2w1sdSJu3GVeHCDfy3mjq8RkzkN3Vh7Di3cB8NRzkwkLQ2FAjxX1kvkNdP9hNBzyBVEJdeWwyb5GfFYXgKe7rPgvWj2QD8FE4W3'
+        address = get_address(self.client, parse_path(path)).address
```
matejcik: ISTM this will fail now that `get_address` returns the address directly?

ddeath: Yes it failed. Fixed. Also replaced custom mnemonic for all all...

trezorlib/tests/device_tests/test_msg_cardano_sign_message.py
ddeath: I can change it as I will be removing / changing all the tests because of change of derivation scheme in the next PR.

matejcik:
In the meantime, I merged a big change into master. This should affect you very little, although you will probably get merge conflicts in `trezorctl`.

IMHO the safest way to resolve them is simply copy-paste your code from pre-merge `trezorctl` to post-merge `trezorctl`.

Another change that affects you is that `@expect` and `@field` are now the same decorator, so please modify that one use of `@field` to `@expect(..., field='...')`

afterwards, run `make style` (note you need python 3.6 for that)

ddeath:
@matejcik should be fixed now

tsusanka:
> @tsusanka please ignore last conversation about i vs i'. It should be ok as it is, because as stated here trezor/trezor-crypto#179 (comment) the derivation scheme v1 is going to be dropped

Ok so `44'/1815'/a'/{0,1}/i` it is.

tsusanka:
IMHO it is correct now and the tests seem to use correct paths.

ddeath:
I am ok with the current paths

ddeath:
@matejcik added changes for cadrano derivation path v2 and marked all test as xfail

matejcik:
Neat, merging. Now let's see what happens with core :)
Thanks!

PR #306, opened 2018-09-06
Add network param to cardano sign tx

ddeath:
so it is possible to switch between mainnet and testnet


prusnak:
https://github.com/trezor/trezor-common/pull/205 is merged

PR #310, opened 2018-09-15
Add Tron support

fbsobreira:
Add Tron support and test cases


matejcik:
Closing for now.

We are moving to a monorepo-based development in https://github.com/trezor/trezor-firmware. When you are ready to continue development on the Tron PR, please add these changes to the same PR.

PR #311, opened 2018-09-18
Stellar - failing test when an asset type of native is explicitly specified

zulucrypto:
This is a failing test for https://github.com/trezor/trezor-core/pull/350


tsusanka:
Fixed style and merged

Issue #307, opened 2018-09-06
Split off UI code from client.py


matejcik:
another important piece of refactoring is moving interaction code into some sort of encapsulated piece away from `client.py`.

Current plan is to create a class for UI interactions, provide two default implementations (text-based and Qt-based), and requiring an instance of this class in `TrezorClient` constructor.

matejcik:
With mixin removal, this is resolved.

PR #314, opened 2018-10-03
WIP: remove TextUiMixin, allow input flows for tests

matejcik:
This implements #307 by moving the UI interaction into a separate module, separated by an actual API.

This change also lets me separate `TrezorClientDebugLink` better and add some useful features for the test suite, namely, _input flow_ support.

In a test, you can write a function that `yield`s to wait for interaction, and then performs the interaction. This is in contrast to the current behavior, where if you want to do anything more complicated than auto-clicking "yes" on every prompt, you need to write send/receive messages by hand, instead of using library functions.

Example usage is in [test_msg_resetdevice_t2.py](https://github.com/trezor/python-trezor/blob/11765498d9b07cad9128606ef6e07d0e904c723b/trezorlib/tests/device_tests/test_msg_resetdevice_t2.py)

In addition, reporting for un/expected messages is nicer.

---

This is work in progress, please do not merge


trezorlib/client.py
```diff
@@ -120,313 +80,85 @@ def __get__(self, instance, cls):
...
-        return proto.ButtonAck()
-
-    def callback_PinMatrixRequest(self, msg):
-        if self.pin_correct:
-            pin = self.debug.read_pin_encoded()
-        else:
-            pin = "444222"
-        return proto.PinMatrixAck(pin=pin)
+        __tracebackhide__ = True  # for pytest # pylint: disable=W0612
+        # do this raw - send ButtonAck first, notify UI later
```
matejcik: ...of course, this breaks Bridge, because of course it does.

tsusanka:
> Example usage is in test_msg_resetdevice_t2.py

So, in a test I define a callable function (`input_flow` in this example) where I specify the confirms/interactions with trezor using `yield` and the debug link and I then give this callback to `self.client.set_input_flow` and run the test. Correct?

LGTM

matejcik:
rebased & merged

PR #315, opened 2018-10-03
Fix Qt5/4 import strategy.

jhoenicke:
- Try Qt5 before Qt4.
- Handle all exceptions (a ValueError is thrown if the wrong Qt was imported earlier)


jhoenicke:
This fixes this bug with electrum:

```
  File "/home/hoenicke/.local/lib/python3.6/site-packages/trezorlib/qt/pinmatrix.py", line 21, in <module>
    from PyQt4.QtGui import (
ValueError: PyCapsule_GetPointer called with incorrect name
```

trezorlib/qt/pinmatrix.py
```diff
@@ -18,34 +18,34 @@
...
         QVBoxLayout,
         QHBoxLayout,
     )
-    from PyQt4.QtCore import QObject, SIGNAL, QRegExp, Qt, QT_VERSION_STR
-except ImportError:
-    from PyQt5.QtWidgets import (
+    from PyQt5.QtGui import QRegExpValidator
+    from PyQt5.QtCore import QRegExp, Qt
+    from PyQt5.Qt import QT_VERSION_STR
+except Exception:
```
prusnak: Why the change from `ImportError` into `Exception`?

matejcik:
looks good, thanks

PR #319, opened 2018-10-09
decred: Return tree and version in insight api

matheusd:
These fields are required when using the python client to correctly send transaction data for signing.


prusnak:
Thanks!

PR #325, opened 2018-10-23
Capricoin support tests and minor trezorctl addition.

tarpas:
This adds tests for https://github.com/trezor/trezor-core/pull/381 .
Please comment/merge.


tarpas:
There is a build (make stylecheck) failure which comes from upstream:
flake8
./trezorlib/tests/device_tests/test_msg_getaddress.py:153:50: F841 local variable 'exc' is assigned to but never used

prusnak:
I pushed codestyle fix. Please rebase on top of current master and squash the commits into one

tarpas:
Done.

trezorlib/tests/device_tests/common.py
```diff
@@ -39,10 +39,13 @@ class TrezorTest:
     pin6 = "789456"
     pin8 = "45678978"
 
+    def coin(self):
+        return "Bitcoin"
+
```
matejcik: let's not do this whole thing

trezorlib/tests/device_tests/common.py
```diff
@@ -39,10 +39,13 @@ class TrezorTest:
...
     def setup_method(self, method):
         wirelink = conftest.get_device()
         self.client = TrezorClientDebugLink(wirelink)
-        self.client.set_tx_api(coins.tx_api["Bitcoin"])
+        self.client.set_tx_api(coins.tx_api[self.coin()])
```
matejcik: let's not do this whole thing

trezorlib/tests/device_tests/test_msg_getaddress.py
```diff
@@ -103,6 +103,15 @@ def test_bch(self):
...
         )
 
+    @pytest.mark.xfail
+    @pytest.mark.capricoin
+    def test_cpc(self):
+        self.setup_mnemonic_allallall()
+        assert (
+            btc.get_address(self.client, "Capricoin", parse_path("m/44'/289'/0'/0/0"))
+            == "CMSgH7wq4kV9ogmSPB5rBmPceQJy3oA9Bu"
+        )
```
matejcik: are Capricoin changes affecting address generation? if not, we can omit this test

trezorlib/tests/device_tests/test_msg_signtx.py
```diff
@@ -64,7 +64,9 @@
 )
 
 
-def check_sign_tx(client, coin_name, inputs, outputs, fee_too_high=False, failure=None):
+def check_sign_tx(
+    client, coin_name, inputs, outputs, fee_too_high=False, failure=None, timestamp=None
+):
```
matejcik: let's not do this whole thing

trezorlib/tests/device_tests/test_msg_signtx_capricoin.py
```diff
@@ -0,0 +1,220 @@
...
+)
+
+TXHASH_8c4553 = bytes.fromhex(
+    "8c4553a62d28a2aa605dc82e80c8e30fdadd49bf950d902c37c34428d5ff58a1"
+)
+
+
+@pytest.mark.xfail
+@pytest.mark.capricoin
+class TestMsgSigntx(TrezorTest):
```
matejcik: This is effectively testing the Bitcoin signing flow except with a timestamp. That's not needed. Please reduce this to one or at most two tests where you provide the timestamp and check that the resulting signature is correct, without involving `check_sign_tx`.

You could also do a fail-test, asserting that signing for Capricoin fails if timestamp is missing.
(Or test for other possible features that work *differently* for Capricoin.)

tarpas: So I should write my own super short/simple version of check_sign_tx, right? Any example I could steal from? If not don't worry I'll manage. (tomorrow/thursday)

tarpas: You're right a single test is sufficient. I was thinking about the fail-test. We could add asserts like below but there is probably not much value.
```python
        with self.client:
            signatures, tx = btc.sign_tx(
                self.client,
                "Capricoin",
                [inp1, inp2],
                [out1],
                version=1,
                timestamp=None,
            )

        assert timestamp_signatures != signatures
        assert len(tx) < len(timestamp_tx)

```
Waiting for your feedback.

matejcik: will this non-timestamped signature be _valid_ for Capricoin?

if yes, then this is a reasonable test.
if not, you should update trezor-core code to throw a failure, and test for that

matejcik: actually, if the timestamp is not mandatory, you should also assert that this creates an expected signature, not just that it's different (you're right about that; checking that the result is different is by itself not interesting)

tarpas: timestamp is mandatory so non-timestamped signature is invalid for Capricoin.

So overwintered (only for Zcash) is optional for Zcash and that's why it doesn't have validation in core / signing.py?

I'll work on new version, thanks.

matejcik: actually we had an internal discussion and let's keep it as it is now, without the fail-test.

it should be there technically, but it is not worth complicating the trezor-core implementation over it

tarpas: Agree with everything :). So I think it's your turn, thanks.

matejcik:
looking good. thanks!

PR #329, opened 2018-10-24
Add coverage for cardano signTx with change address

ddeath:
In current test there is not covered option when transaction has change address.

This PR adding coverage for that.

Fixing https://github.com/trezor/trezor-core/issues/382

### DIFF
--- a/trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
+++ b/trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
@@ -27,7 +27,7 @@ from .common import TrezorTest
 @pytest.mark.cardano
 @pytest.mark.skip_t1  # T1 support is not planned
 class TestMsgCardanoSignTx(TrezorTest):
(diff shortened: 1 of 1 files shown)


prusnak:
I would rather have two distinct test fixtures instead of replacing the old one with the new one.

Can you add a new fixture instead of modifying the old one?

matejcik:
while you're at it, maybe take a look at https://github.com/trezor/python-trezor/blob/1397c3b4d9e6c95e44a0e6460d02b817f73e8492/trezorlib/tests/device_tests/test_msg_resetdevice_t2.py#L37 ?

ddeath:
I made a separate fixture. Tried to put it into `input_flow()` but it was not working so didnt changed it for now.

matejcik:
I refactored your original test in the meantime - apparently the swipe-down is not needed anymore?

Please look at the new version and check if it can work with your new test.

matejcik:
superseded by #342

PR #336, opened 2018-11-06
Add Eos support

atkachyshyn:


prusnak:
Depends on https://github.com/trezor/trezor-common/pull/231

prusnak:
Depends on https://github.com/trezor/trezor-common/pull/231

matejcik:
why is there a truckload of JSON files in a new `misc` directory?

atkachyshyn:
Its the examples of the all possible transactions to be signed, so anybody
can check them right away.
Should i remove them or probably move to some other place?



ср, 7 лист. 2018 о 14:42 matejcik <notifications@github.com> пише:

matejcik:
You could turn them into test cases. That's missing from your submission anyway. In that case they should move into `trezorlib/tests/device_tests/eos` or something like that.

atkachyshyn:
OK, thanks for the suggestion. Will do that.

ср, 7 лист. 2018 о 15:03 matejcik <notifications@github.com> пише:

trezorlib/eos.py
```diff
@@ -0,0 +1,387 @@
...
+
+
+
+def parse_transaction_json(json):
+    tx = type('Transaction', (object,), {})()
+    tx.chain_id = binascii.unhexlify(json['chain_id'])
+
+    body = json['transaction']
+
+    expiration = int(datetime.strptime(body['expiration'], '%Y-%m-%dT%H:%M:%S').strftime("%s"))
```
keepkeyjon: IMO, it would be better if this didn't adjust to the local timezone:

```suggestion
    expiration = int((datetime.strptime(body['expiration'], '%Y-%m-%dT%H:%M:%S') - datetime(1970, 1, 1)).total_seconds())
```

keepkeyjon: In fact, it's necessary to make the testcases timezone agnostic. As-is, these only work in UTC+3.

matejcik (changes requested):

matejcik:
closing for cleanup now.
please apply your changes to the monorepo at https://github.com/trezor/trezor-firmware, together with the corresponding trezor-core PR, and resubmit.

also please update your python-trezor code to match the current implementation, esp. the testcases.

please make use of `trezorlib.protobuf.dict_to_proto` where appropriate, instead of having explicit converters for the json fields.

Issue #332, opened 2018-10-31
Add a test case for segwit inputs/outputs with very high amounts


jpochyla:
Implement a sign_tx test that covers this bug: https://github.com/trezor/trezor-core/issues/389

PR #337, opened 2018-11-08
Add a test case for segwit inputs/outputs with very high amounts

strmci:
Add a test case for segwit inputs/outputs with very high amount, solves issue #332

### DIFF
--- a/trezorlib/tests/device_tests/test_msg_signtx.py
+++ b/trezorlib/tests/device_tests/test_msg_signtx.py
@@ -63,6 +63,12 @@ TXHASH_50f6f1 = bytes.fromhex(
     "50f6f1209ca92d7359564be803cb2c932cde7d370f7cee50fd1fad6790f6206d"
 )
 
+TXHASH_fffff = bytes.fromhex(
+    "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
+)
+
+TX_API = tx_cache("Testnet")
+
 
 def check_sign_tx(client, coin_name, inputs, outputs, fee_too_high=False, failure=None):
     __tracebackhide__ = True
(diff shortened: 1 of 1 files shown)


matejcik:
you will need to run `make style` to fix the style warnings in Travis

given that this is a SegWit test, you should move it to `test_msg_signtx_segwit.py`

Maybe it's worth it to have a non-segwit test for the same thing too, using the transactions you found earlier? I'm not sure if Trezor will let you sign that though.

trezorlib/tests/device_tests/test_msg_signtx.py
```diff
@@ -63,6 +63,12 @@
     "50f6f1209ca92d7359564be803cb2c932cde7d370f7cee50fd1fad6790f6206d"
 )
 
+TXHASH_fffff = bytes.fromhex(
```
matejcik: this chunk should not be here. you don't need `TX_API` and see below for txhash

trezorlib/tests/device_tests/test_msg_signtx.py
```diff
@@ -309,6 +315,63 @@ def test_one_three_fee(self):
...
 
+    def test_testnet_segwit_big_amount(self):
+        self.setup_mnemonic_allallall()
+
+        inp1 = proto.TxInputType(
+            address_n=parse_path(
+                "m/49'/1'/0'/0/0"
+            ),
+            amount=2**32+1,
+            prev_hash=TXHASH_fffff,
```
matejcik: instead:
```python
prev_hash=b"\xff" * 32
```

trezorlib/tests/device_tests/test_msg_signtx.py
```diff
@@ -309,6 +315,63 @@ def test_one_three_fee(self):
...
+            prev_index=0,
+            script_type=proto.InputScriptType.SPENDP2SHWITNESS,
+
+        )
+
+        out1 = proto.TxOutputType(
+            address="2N1gtqietBUnAb5aaTupRoJt1vqFuozL9cY",
```
matejcik: there should be a comment explaining where the address comes from

trezorlib/tests/device_tests/test_msg_signtx.py
```diff
@@ -309,6 +315,63 @@ def test_one_three_fee(self):
...
+                    ),
+                    proto.TxRequest(
+                        request_type=proto.RequestType.TXINPUT,
+                        details=proto.TxRequestDetailsType(request_index=0),
+                    ),
+                    proto.TxRequest(request_type=proto.RequestType.TXFINISHED),
+                ]
+            )
+            _, serialized_tx = btc.sign_tx(
+                self.client, "Testnet", [inp1], [out1], prev_txes=TX_API
```
matejcik: you should not need the `prev_txes` argument

trezorlib/tests/device_tests/test_msg_signtx.py
```diff
@@ -508,6 +511,31 @@ def test_p2sh(self):
...
+            amount=411102528330,
+            prev_hash=TXHASH_2bac7a,
+            prev_index=1,
+        )
+        out1 = proto.TxOutputType(
+            address="2N1gtqietBUnAb5aaTupRoJt1vqFuozL9cY",  # Bitcoin Testnet address (randomly chosen)
+            amount=411102528330,
+            script_type=proto.OutputScriptType.PAYTOADDRESS,
+        )
+        _, serialized_tx = btc.sign_tx(
```
matejcik: now it should be possible to use `_check_sign_tx` here

PR #339, opened 2018-11-28
add input flow function to NEM signing test

strmci:


matejcik:
merged via b0a227866ad4fe4320720422d81ae8caa2999c9f
thanks!
now if you could do this for the remaining uses of swipes too (they're in Ontology tests)

trezorlib/tests/device_tests/test_msg_nem_signtx_mosaics_t2.py
```diff
@@ -168,33 +169,38 @@ def test_nem_signtx_mosaic_creation_levy(self):
...
-        self.client.debug.press_yes()
-        self.client.transport.write(proto.ButtonAck())
-        ret = self.client.transport.read()
-
-        # Confirm tx
-        assert isinstance(ret, proto.ButtonRequest)
-        self.client.debug.press_yes()
-        self.client.transport.write(proto.ButtonAck())
-        return self.client.transport.read()
+            for i in range(num_of_swipes):
```
matejcik: one very minor thing that I missed:

if the `i` is unused in the loop, please use `_` for the variable, as in:
```python
for _ in range(num_of_swipes):
	do.stuff()
```

PR #342, opened 2018-11-30
Change network to protocol magic in cardano

refi93:
Related to https://github.com/trezor/trezor-core/pull/417

### DIFF
--- a/trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
+++ b/trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
@@ -14,6 +14,8 @@
 # You should have received a copy of the License along with this library.
 # If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.
 
+import time
+
 import pytest
 
 from trezorlib import cardano, messages
@@ -21,10 +23,10 @@ from trezorlib import cardano, messages
 from .conftest import setup_client
 
 VECTORS = [
-    # Mainnet transaction
+    # Mainnet transaction without change
     (
-        # network
-        2,
+        # protocol magic (mainnet)
+        764824073,
         # inputs
         [
             {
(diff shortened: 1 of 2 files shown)


matejcik:
please wait until the trezor-common PR is accepted and then update the reference to trezor-common submodule

refi93:
@jpochyla I tried adding to `test_cardano_sign_tx` function the following swipe and button press event emulations, following your suggestion in https://github.com/trezor/trezor-core/pull/417#issuecomment-454422223:
```
    client.debug.swipe_down()
    client.debug.press_yes()
    client.debug.swipe_down()
    client.debug.press_yes()
```
and Trezor during this test swipes only through the first screen (confirm address), confirms the address and hangs on the second screen, waiting for the second swipe, although I have there the second swipe_down instruction. Seems like the second swipe down and press_yes events happen prematurely. I tried even adding time.sleep(1) in between them, as in other examples in the code, without success, which I kind of expected and understand, since the sleep happens before sending the transaction to Trezor. So how to properly make the second swipe and "hold to confirm"?

matejcik:
> @jpochyla I tried adding to `test_cardano_sign_tx` function ([trezor/python-trezor/pull/342/files#diff-ecde254718349399eb838e55d428d744L89](https://github.com/trezor/python-trezor/pull/342/files#diff-ecde254718349399eb838e55d428d744L89)) the following swipe and button press event emulations, following your suggestion in [trezor/trezor-core#417 (comment)](https://github.com/trezor/trezor-core/pull/417#issuecomment-454422223):
> 
> ```
>     ...
>     with client:
>         client.set_expected_responses(expected_responses)
>         client.debug.swipe_down()
>         client.debug.press_yes()
>         client.debug.swipe_down()
>         client.debug.press_yes()
> ```

that's not at all what [the referenced code](https://github.com/trezor/python-trezor/blob/master/trezorlib/tests/device_tests/test_msg_resetdevice_t2.py#L56) does though?

you need to put the input interactions into a separate "input flow" function, set it up as a receiver with `client.set_input_flow`, and use `yield` statements to wait for `ButtonRequest`s instead of spamming all input instructions in one go

refi93:
@matejcik I see, thanks. So I tried it this way:
```
    ...
    def input_flow():
        client.debug.swipe_down()
        client.debug.press_yes()
        
        yield

        client.debug.swipe_down()
        client.debug.press_yes()

    with client:
        client.set_expected_responses(expected_responses)
        client.set_input_flow(input_flow)
        response = cardano.sign_tx(
            client, inputs, outputs, transactions, protocol_magic
        )
        assert response.tx_hash.hex() == tx_hash
        assert response.tx_body.hex() == tx_body
```
and Trezor still hangs on the second screen. It seems as if the control is never returned to input_flow() after the yield statement. I tried also adding sleep statements between the input instructions without success. What am I missing? To be honest, I don't have much experience with Python, particularly with generator functions, therefore I can be doing a dumb mistake somewhere, so don't hesitate to point it out. Thanks!

matejcik:
Most likely you need to `yield` at start of the `input_flow()` function; otherwise, the first two input calls run immediately.
In general, each `yield` translates to "wait until ButtonRequest", so there should be as many `yield`s as there are expected `ButtonRequest`s, which in your case is 4?

refi93:
yes, I tried putting a `yield` at the beginning of the input_flow() function, but it also seems to hang and it doesn't get past that yield (tried throwing an exception right after the yield, which was not thrown). It seems as if no ButtonRequests were coming from Trezor

matejcik:
that could be the case. run the test with `pytest -v -s` to see messages as they come and go.
i might try and give your PR a go later. this needs https://github.com/trezor/trezor-core/pull/417 to work, correct?

refi93:
yes, and trezor-common https://github.com/trezor/trezor-common/pull/244 , since there were changes to the message definitions

prusnak:
I just merged https://github.com/trezor/trezor-common/pull/244 into master

matejcik:
well, your code doesn't seem to be _sending_ the ButtonRequests, so that's a problem right there.

refi93:
@matejcik thanks! I just added them to https://github.com/trezor/trezor-core/pull/417 and it works

refi93:
moreover, I added the test case from this unmerged PR: https://github.com/trezor/python-trezor/pull/329 so it can be closed after approving and merging this one

refi93:
I recently added tests for cardano address validation - a recent issue we came across

trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
```diff
@@ -150,3 +229,32 @@ def input_flow():
...
+    inputs = [cardano.create_input(i) for i in inputs]
+    outputs = [cardano.create_output(o) for o in outputs]
+
+    expected_responses = [
+        messages.CardanoTxRequest(tx_index=i) for i in range(len(transactions))
+    ]
+    expected_responses += [messages.Failure()]
+
+    with client:
+
```
matejcik: less whitespace please :))

trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
```diff
@@ -150,3 +229,32 @@ def input_flow():
...
+        client.set_expected_responses(expected_responses)
+
+        try:
+            cardano.sign_tx(client, inputs, outputs, transactions, protocol_magic)
```
matejcik: please use `with pytest.raises` instead of try-fail

matejcik: see https://github.com/trezor/python-trezor/blob/master/trezorlib/tests/device_tests/test_msg_signtx_bcash.py#L337 for usage

trezorlib/tests/device_tests/test_msg_cardano_sign_transaction.py
refi93: seems to be introduced by "make style" when I run it

PR #344, opened 2018-12-06
blockbook missing locktime

brianddk:
blockbook doesn't seem to always return locktime which breaks PAYTOADDRESS in versions < 0.11.  Defaulting to zero for cross-compatibility with existing blockbook servers.

BTW, this fires from `trezorctl sign-tx` if that wasn't clear.


matejcik:
This is fixed in 0.11 which was just released.

Users of trezorctl should upgrade to 0.11 and I'm not aware of other applications that would require 0.10.x and rely on the built-in TxApi, so I'm going to close this.

brianddk:
> I'm not aware of other applications that would require 0.10.x

I believe that Electrum 3.2.2 still requires 0.10.x (ie its not forward compatible).  Other conflicts prevent me from testing Electrum@master but I might later.  Plus all the Electrum forks (BCH, LTC, FTC.. etc) are quite a few commits behind Electrum@master so they are even deeper in the whole.

I've patched my install since I'm not really using `locktime` or LN in any of my TXNs presently.

PR #346, opened 2018-12-06
Windows instal doc

brianddk:
Suggest steering windows users to choco to satisfy the protoc and Visual C++ dependencies


matejcik:
thanks :)

PR #349, opened 2018-12-26
qt: don't import PyQt5.Qt

SomberNight:
Apparently, `import PyQt5.Qt` imports "all of Qt".
Let's not do this please. :)

- we have "bluetooth"-related issues
- binaries created by pyinstaller are unnecessarily large

see https://github.com/spesmilo/electrum/issues/4960

### DIFF
--- a/trezorlib/qt/pinmatrix.py
+++ b/trezorlib/qt/pinmatrix.py
@@ -30,8 +30,7 @@ try:
         QHBoxLayout,
     )
     from PyQt5.QtGui import QRegExpValidator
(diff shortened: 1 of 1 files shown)


prusnak:
Seems legit; @matejcik what do you think? Please merge if you are OK with the change.

matejcik:
agreed, I'll merge & release tomorrow

matejcik:
I have merged this. I'm holding off on releasing because a Travis build got broken and while I *don't think* it's related, better safe than sorry.

SomberNight:
Thank you for the prompt response

PR #351, opened 2018-12-30
Dash: Support spending DIP2 special transaction inputs

codablock:
This implements support for spending of TX outputs which are part of a [DIP2](https://github.com/dashpay/dips/blob/master/dip-0002.md) special transaction, especially [DIP4 coinbases](https://github.com/dashpay/dips/blob/master/dip-0004.md#coinbase-special-transaction).

This is for Dash only and thus required the addition of `is_dash` in `tx_api.py`.

Support on the firmware side is not required for this to work as it reuses the logic from Zcash's `extra_data` and `extra_data_len` fields.

This is currently not easily testable as it requires a version of insight-api that supports pass through of DIP2 payloads, which is not deployed on our testnet insight instance yet. I'll leave a comment here when this is done.


codablock:
Don't know where to ask best, so I'm asking here. Can someone help me figuring out where to start with integrating the same changes to wallet.trezor.io? I already found the https://github.com/trezor/trezor-wallet repo, but this seems to not be the same wallet that is running on wallet.trezor.io and also seems to have no Dash support yet (selecting Dash just reloads the page and does nothing more). @prusnak maybe?

prusnak:
@codablock the source of the wallet running on wallet.trezor.io is not public, only the upcoming version has public source

codablock:
@prusnak Thanks for the info. So I assume I can't really help on source level. Anything else I can do to help, e.g. creating an issue somewhere?

prusnak:
@codablock You can help by adding this functionality to https://github.com/BitGo/bitgo-utxo-lib/ library

trezorlib/tx_api.py
```diff
@@ -104,6 +109,30 @@ def json_to_tx(coin, data):
             extra_data_len = 1 + joinsplit_cnt * 1802 + 32 + 64
             t.extra_data = rawtx[-extra_data_len:]
 
+    if is_dash(coin):
+        t.type = data["type"]
+        if t.version == 3 and t.type != 1:
+            # It's a DIP2 special TX with payload
+            t.extra_data_len = data["extraPayloadSize"]
```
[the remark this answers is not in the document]
matejcik: also `extra_data_len` is recalculated anyway so you don't need to fill it out, you only need the length for the following varint

trezorlib/tx_api.py
```diff
@@ -104,6 +109,30 @@ def json_to_tx(coin, data):
...
+            if t.extra_data_len == 0:
+                t.extra_data_len = 1
+                t.extra_data = b"\x00"
+            elif t.extra_data_len < 253:
```
matejcik: please extract this code into a separate function - something like `pack_varint(num)` which returns a `bytes` object

(if you could then use that function also in the zcash code, that would be extra nice .) )

alternately, if viable (how likely is it that the payload is more than 253 bytes?), you could be lazy, do the thing that ZCash does and raise if length is 253 or more

trezorlib/tx_api.py
```diff
@@ -104,6 +109,30 @@ def json_to_tx(coin, data):
...
+                t.extra_data = struct.pack("<B", t.extra_data_len)
+                t.extra_data_len += 1
+            elif t.extra_data_len <= 65535:
+                t.extra_data = struct.pack("<BH", 253, t.extra_data_len)
+                t.extra_data_len += 3
+            else:
+                t.extra_data = struct.pack("<BI", 254, t.extra_data_len)
```
matejcik: use `<BL` instead

also please add the `<BQ` case

trezorlib/tx_api.py
codablock: It's actually a viable scenario that we get >= 253 bytes for the payload. Depends on which DIP2 types we later introduce. I'll extract the serialization code into its own method.

codablock: I tried to use the freshly introduces `pack_varint` for ZCash as well, but then decided to not touch that part as it's not a simply replacement anymore. I would also have to change the way the data is extracted from `rawtx` and even though this is a small change, I find it to risky to touch code that I can't test/verify to work afterwards.

trezorlib/tx_api.py
```diff
@@ -104,6 +109,30 @@ def json_to_tx(coin, data):
             extra_data_len = 1 + joinsplit_cnt * 1802 + 32 + 64
             t.extra_data = rawtx[-extra_data_len:]
 
+    if is_dash(coin):
+        t.type = data["type"]
```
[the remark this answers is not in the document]
codablock: Changed to use a local/temporary variable only (Also verified that we actually don't need the type field later)

trezorlib/tx_api.py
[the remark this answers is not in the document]
codablock: Added code to verify existence of extraPayloadSize/extraPayload and correctness of extraPayloadSize. Also removed assignment of extra_data_len and only using local var now.

trezorlib/tx_api.py
codablock: Done (in `pack_varint`)

codablock:
Handled requested changes and force-pushed new version.

trezorlib/tx_api.py
```diff
@@ -32,6 +33,23 @@ def is_capricoin(coin):
...
+
+
+def pack_varint(n):
+    if n == 0:
+        return b"\x00"
+    elif n < 253:
+        return struct.pack("<B", n)
+    elif n <= 65535:
+        return struct.pack("<BH", 253, n)
+    elif n <= 4294967295:
```
matejcik: please use hex constants here (`0xFFFF`, `0xFFFFFFFF`). it makes it clearer on first look that the constants are correct

trezorlib/tx_api.py
```diff
@@ -32,6 +33,23 @@ def is_capricoin(coin):
     return coin["coin_name"].lower().startswith("capricoin")
 
 
+def is_dash(coin):
+    return coin["coin_name"].lower().startswith("dash")
+
+
+def pack_varint(n):
+    if n == 0:
+        return b"\x00"
```
matejcik: this case can be folded into the `n < 253` case

trezorlib/tx_api.py
```diff
@@ -104,6 +122,28 @@ def json_to_tx(coin, data):
...
+        if t.version == 3 and dip2_type != 0:
+            # It's a DIP2 special TX with payload
+
+            if "extraPayloadSize" not in data or "extraPayload" not in data:
+                raise ValueError("Payload data missing in DIP2 transaction")
+
+            if data["extraPayloadSize"] * 2 != len(data["extraPayload"]):
+                raise ValueError(
+                    "extra_data_len (%d) does not match calculated length (%d)"
+                    % (data["extraPayloadSize"], len(data["extraPayload"]) * 2)
```
matejcik: should be  `len() / 2`
but maybe better to just throw out the detailed info and leave the ValueError at "length mismatch"

codablock:
@matejcik I think I handled all the review comments now. Also added device tests and 2 more unit tests (for pre-DIP2 TXs). Also added a fix for a KeyError when accessing `data["type"]` which might happen for pre-DIP2 TXs.

matejcik:
looking good, thanks

PR #352, opened 2019-01-04
Don't require prevtx for coins with BIP-143 fork id

jhoenicke:
Is using `coins.by_name` okay here, to check for `force_bip143`?

This is necessary to get `electron-cash` working again (plus the obvious update of the plugin in electroncash – I'm working on that).


jhoenicke:
I added the `if coin_name in coins.by_name` you suggested, just fixed the polarity (load_prevtxes is true if force_bip143 is false).

I also added the `defaultdict` workaround to electroncash so that it works with trezor 0.11.1.

matejcik:
merged, thanks!

PR #353, opened 2019-01-11
Add get public key to Ethereum

tsusanka:


tsusanka:
Since this functionality is not yet implemented in trezor-core we should probably wait. Tracked in https://github.com/trezor/trezor-core/issues/406

matejcik:
does trezor-mcu have it?

tsusanka:
It does not, I'm sorry I keep forgetting about mcu.

What is the correct approach to this? Do we have a tag or smth like that to denote something is missing on T1?

I could implement this on trezor-mcu, but I'm not sure it is worth it, since we plan to deprecate it. In the meanwhile the frontend can use the Bitcoin's get public key as it does currently.

matejcik:
that's probably more up to wallet devs to decide; this way they'll have a non-uniform interface

anyway, you'd have to mark the test as `@pytest.mark.skip_t1`

tsusanka:
Agreed to be implemented. PR in https://github.com/trezor/trezor-mcu/pull/447 . So this should wait for it

prusnak:
Just merged into trezor-mcu

matejcik:
merged, thanks

matejcik:
ah darn, should've waited for the trezor-core PR. oh well.

PR #354, opened 2019-01-17
Makefile: typo

tsusanka:
Just a rename from `make stylecheck` to `make style_check` to make it consistent with trezor-core


matejcik:
thanks :)

PR #358, opened 2019-02-11
Add support for custom explorers

AlexITC:
Now, the  `custom_explorer` field could be added to the json files, for example:
```
    "custom_explorer": {
      "api": "stakenet.StakenetTxApi",
      "urls": [
        "https://xsnexplorer.io"
      ]
    }
```

Then, the api class is loaded dynamically, which allows to decouple from the blockbook explorer.

There are some possible improvements:
- Remove method duplication.
- Use inheritance to enforce the api contract.
- Possibly move the original `tx_api.py` to the `explorers` module.

The goal of this PR is to discuss a bit about what do you guys expect and make the necessary changes after that.

Thanks.

### DIFF
--- a/trezorlib/coins.py
+++ b/trezorlib/coins.py
@@ -44,8 +44,28 @@ except Exception as e:
     raise ImportError("Failed to load coins.json. Check your installation.") from e
 
 slip44 = {name: coin["slip44"] for name, coin in by_name.items()}
-tx_api = {
+blockbook_tx_api = {
     name: TxApi(coin)
     for name, coin in by_name.items()
     if coin["blockbook"] or coin["bitcore"]
 }
+
+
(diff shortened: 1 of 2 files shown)


AlexITC:
Travis fixed.

matejcik:
explorer-related code should not be in python-trezor at all, its presence is a historical vestige. I'm going to remove it in a matter of months, so there's no point in improving it.

Maybe your code will be useful for microwallet, a new project that's supposed to take over this part of trezorctl's responsibilities. But the problem is that we don't really want to be stuck maintaining a bunch of different explorer APIs. I'll try to figure out some sort of plugin architecture, possibly based on setuptools entry points, so that it could be extended outside the base distribution.

PR #362, opened 2019-02-24
Added a few more sample scripts

brianddk:
### encrypt_value.py
Use trezor to encrypt an arbitrary value from stdin and armor
the encrypted value to stdout.  Padding and length checks are
automatically handled.  Key and Address are generated or taken
from command line

### offline_txn.py
A very simple offline transaction assuming blockbook TXN json
is provided by the user.  A good way sign a known transaction
on an air-gapped system

### simple_txn.py
A very simple online transaction.

### wallet_search.py
A script to walk trezor coins and addresses looking for
activity.  This only works on coins with blockbook servers.

### DIFF
--- a/tools/encrypt_value.py
+++ b/tools/encrypt_value.py
@@ -2,33 +2,37 @@
 #
 # Note, this doesn't use the same address (-n) as Trezor Password Manager
 #   By default it will store the address and key in clear text and only
-#   encrypt the value.  Use --secure to force the user to name key and 
-#   address on the command line, and keep it out of the armor file. 
+#   encrypt the value.  Use --secure to force the user to name key and
+#   address on the command line, and keep it out of the armor file.
 
 from trezorlib import misc, tools, ui
 from trezorlib import __version__ as trezor_version
 from trezorlib.client import TrezorClient
 from trezorlib.transport import get_transport
 from random import getrandbits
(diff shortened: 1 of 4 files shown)


matejcik:
**encrypt_value.py** - I don't think this is too useful, given that you can encrypt files in a more-or-less standardized manner with [GPG](https://github.com/romanz/trezor-agent/blob/master/doc/README-GPG.md)

**wallet_search.py** - This is not the right repo for that kind of script. For one, people would use it and then we'd be forced to fix bugs in it :) Also it's going to be partially superseded by the upcoming microwallet tool.
Besides, it's not a great example script. You regenerate a client connection for every coin, load JSON data manually (that might not be present on the system that runs it), etc.
I'd recommend keeping it in your own repo or in a gist.

---

That leaves **simple_txn.py** and **offline_txn.py**, which are neat example scripts. I've left review comments on offline_txn. I'd actually recommend merging them into one script, because the `tx_api` interface will be removed soon and so simple_txn will stop working.

matejcik (changes requested):
ha, i _didn't_ submit the reviews apparently.
here goes

matejcik:
closing for now.
whenever you want to continue on this, please open a PR against https://github.com/trezor/trezor-firmware

PR #363, opened 2019-02-27
Fix incompatibilities with v0.11 movement of methods to `misc` library.

brianddk:
### pwd_reader.py

* Moved encrypt/decrypt calls to misc module from client.
* Corrected key to encode in binstr for Py3.7 hmac functions.
* Added ui to client.

### pwd_reader.py

* Moved entropy calls to misc module from client.
* Added ui to client.


matejcik:
neat, thanks

PR #364, opened 2019-03-01
Added ClickUI, useEmptyPassphrase, and try clauses for bridge / webusb

brianddk:
* mem_flashblock.py, mem_read.py, mem_write.py
  * Added ClickUI to remove warning messages on client instantiation.
  * Added try / except for systems with both bridge and webusb devices.
* pwd_reader.py, encfs_aes_getpass.py
  * Emulated `useEmptyPassphrase` so that encrypt / decrypt operations will match those of Trezor Password Manager.

### DIFF
--- a/tools/mem_write.py
+++ b/tools/mem_write.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+from trezorlib.ui import ClickUI
 from trezorlib.debuglink import DebugLink
 from trezorlib.client import TrezorClient
 from trezorlib.transport import enumerate_devices
@@ -9,17 +10,28 @@ def main():
     # List all debuggable TREZORs
     devices = [device for device in enumerate_devices() if hasattr(device, 'find_debug')]
 
+    # Try to get debug transport on devices
+    debug_transport = None
+    for dev in devices:
+        try:
+            transport = dev
+            debug_transport = dev.find_debug()
+            break
+        except Exception:
+            pass
+
     # Check whether we found any
     if len(devices) == 0:
         print('No TREZOR found')
         return
 
-    # Use first connected device
-    transport = devices[0]
-    debug_transport = devices[0].find_debug()
+    # Check if its available
+    if not debug_transport:
+        print('Debug device not available')
+        return
 
     # Creates object for manipulating TREZOR
-    client = TrezorClient(transport)
+    client = TrezorClient(transport, ui=ClickUI())
     debug = DebugLink(debug_transport)
 
     debug.memory_write(int(sys.argv[1], 16), bytes.fromhex(sys.argv[2]), flash=True)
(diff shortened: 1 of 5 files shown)


matejcik:
i'm pretty sure the NoPassphraseUI in `encfs_aes_getpass` is wrong. we never disallowed passphrases for encfs

tools/mem_write.py
```diff
@@ -9,17 +10,28 @@ def main():
...
     # Check whether we found any
     if len(devices) == 0:
         print('No TREZOR found')
         return
 
-    # Use first connected device
-    transport = devices[0]
-    debug_transport = devices[0].find_debug()
+    # Check if its available
+    if not debug_transport:
```
matejcik: is there any case where you don't get an exception from the debug device, but get a falsey value for `find_debug`?

brianddk:
> i'm pretty sure the NoPassphraseUI in `encfs_aes_getpass` is wrong. we never disallowed passphrases for encfs

Sure, I'll back it out.  I just referenced the Trezor Password Manager, but didn't look at the storage provider.  I'll make the change.

tools/mem_write.py
brianddk: > is there any case where you don't get an exception from the debug device, but get a falsey value for `find_debug`?

I always either get a valid handle or `TransportException`, but I never get `False`
```
bridge:lib02
webusb:001:1:2
```
Any call to `device['webusb'].find_debug()` raises:
```
TransportException: Debug device not available
```

As long as the enumerator grabs bridge device fist, everything works, but if it happens upon the webusb first, it fails.  This is what happened when I originally made the PR.  I recently noticed that `trezord.exe` had abended, so perhaps this is what I was seeing.

I could look at linux behaviour later, perhaps it is more resilient.

matejcik: oooh I understand what you meant now. No, this is not a good way to do that. These scripts are really old.

Instead do something like this:

```python
def get_debug_device():
	for device in enumerate_devices():
		try:
			return TrezorClientDebugLink(device)
		except Exception:
			pass
	raise Exception("no debuggable device found")

def main():
	# ....
	client = get_debug_device()
	client.debug.memory_write(...)
```

brianddk: Will do... I'll update it in the next day or so.

Thx

brianddk: Does `DebugLink` require debug firmware loaded on the device?  I can't seem to get any `DebugLink` method to run.

I [collected logs](https://gist.github.com/brianddk/a735927014e4024b10edc4efacd01026) for running `py.test trezorlib/tests/device_tests/test_msg_getaddress.py` compared to `py.test trezorlib/tests/device_tests/test_msg_getaddress.py`.  The `py.test` run was failing in DebugLink.

matejcik: yes, debuglink requires a debug firmware.
pytest should be able to run on a debugless firmware if you `export INTERACT=1`

matejcik:
superseded by 2813522b05cef4e0e545a101f8b3559a3183b45b

PR #365, opened 2019-03-02
Fix trezorctl for protobuf changes in 0.11.2

brianddk:
* **ethereum-sign-message** - Leave `address` as text instead of encoding.
* **ethereum-verify-message** - Leave `address` as text instead of encoding.
* **ethereum-sign-tx** - Leave `to` as text instead of encoding.

***NOTE***: This change will break backward compatibility for older FW with
        these commands.

### DIFF
--- a/trezorctl
+++ b/trezorctl
@@ -1041,7 +1041,7 @@ def ethereum_sign_message(connect, address, message):
     ret = ethereum.sign_message(client, address_n, message)
     output = {
         "message": message,
-        "address": "0x%s" % ret.address.hex(),
+        "address": ret.address,
         "signature": "0x%s" % ret.signature.hex(),
     }
     return output
@@ -1060,7 +1060,6 @@ def ethereum_decode_hex(value):
 @click.argument("message")
 @click.pass_obj
 def ethereum_verify_message(connect, address, signature, message):
-    address = ethereum_decode_hex(address)
     signature = ethereum_decode_hex(signature)
     return ethereum.verify_message(connect(), address, signature, message)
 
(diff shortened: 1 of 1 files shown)


matejcik:
thanks
maybe we should bump the minimum required firmware then, to 1.8.0 and 2.1.0. care to add that to the PR? the values are in `trezorlib/__init__.py`.

we probably won't be releasing a new version soon, and when it gets released, the up-to-date versions are hopefully going to be widespread enough

brianddk:
> thanks
> maybe we should bump the minimum required firmware then, to 1.8.0 and 2.1.0. care to add that to the PR? the values are in `trezorlib/__init__.py`.

Sure.  Since I didn't know your release schedule, I didn't initially bump the minimum version.  I'll bump it, but also wanted to check to other protobuf changes.  There were a few other messages I didn't get around to validating.

I'll try to work them in as well.

brianddk:
@matejcik is there any way to downgrade the bootloader and FW from 1.8.0?  I wanted to go back to 1.7.3 or earlier to verify the `MINIMUM_FIRMWARE_VERSION` settings side effects, but now, unfortunately, I'm locked in at 1.8.0 on both devices.

matejcik:
that's not possible.
probably the easiest way to verify the effect is bump up the min version to 1.8.1 though

matejcik:
i'm now working on ethereum subcommands, so i'm merging this now.

thanks for your work!

btw, the backwards compatibility was already broken with the protobuf update, so the changes in trezorctl do nothing worse

PR #369, opened 2019-03-17
Fix TX fetching

yura-pakhuchiy:
```
TypeError: Object of type 'Decimal' is not JSON serializable
```


matejcik:
I'm pretty sure that changing _fetch_ format is a wrong place to fix a problem that occurs when _serializing_.

How did you encounter this error?

yura-pakhuchiy:
Just delete any prefetched transaction from txcache and run test which uses it. The code will fail to save it after fetching and test will fail.

matejcik:
closing, as this part will be reworked anyway.

